PR #10, opened 2020-02-09
code-server: docker initial release

aptalca:
External trigger for docker-compose: https://ci.linuxserver.io/job/External-Triggers/job/code-server-docker-external-trigger/
The trigger could use a review as well.


thelamer:
Does this work ? I thought we needed a single layer to consume , I never tested the maintainer label stuff but I am pretty sure those two copies means you will only get the compose bin

aptalca:
No, you're right. I messed up. Forgot to move everything to root layer and then back in in the last stage like the Dockerfile.complex example.

The maintainer label is just for reference in case someone inspects the mod image

aptalca:
Force pushed single layer image version. The maintainer label didn't affect the other mods I pushed.

PR #11, opened 2020-02-09
code-server-docker fix permissions

aptalca:
fix permissions if docker group on the host is not the default `999`


thelamer:
For Taisun I shifted an existing group if conflict to an unused gid https://github.com/linuxserver/docker-taisun/blob/master/root/etc/services.d/taisun/run

Not sure if that is still relevant

aptalca:
Hmm, wouldn't that get messed up if docker.sock is owned by root? Then it would try to swap root to 65500?

PR #18, opened 2020-03-21
Added mod for ssh tunneling

Dobli:
This mod enables tcp forwarding for the openssh server, allowing to use the container as tunnel, e.g. between a private network and a public server to make local services publicly available. Usage example is provided within the readme.


aptalca:
@Dobli are you sure `expose` is needed? As far as I know, `expose` is only for documentation purposes and has no effect on the container: https://docs.docker.com/engine/reference/builder/#expose

Dobli:
Hi, I'll check it again later, at least when I tried it last time it seemed necessary so my webserver could reach it. I'll try to omitt it and see if it is necessary.

Dobli:
@aptalca 
Okay I checked it again. It behaves like described in my readme. When `exposed` is used in the compose file (or run command) other containers (in my case my reverse proxy) are able to access the port, when I omit it the port is not reachable.

This also matches the description of `expose` in the compose documentation: https://docs.docker.com/compose/compose-file/#expose

You are right that it usually is not necessary, but this could be caused by `docker run` and `docker-compose` commands both respecting the EXPOSE entries within Dockerfiles and exposing these ports to other containers by default: https://docs.docker.com/compose/compose-file/#service-configuration-reference

In this case though, it is unknown what ports should be exposed during Dockerfile creation. Therefore the expose entry is necessary.

aptalca:
Great, thanks for confirming.

PR #37, opened 2020-05-09
add ffmpeg mod to the list

mKeRix:
The mod was created as a response to linuxserver/docker-lazylibrarian#45. As suggested by @CHBMB I'm adding it to the list. If you want to create a new branch in this repo and bring the mod under your control I can also submit a new PR with that.

EDIT: If I should create a new mod I would appreciate it if you could create a new `lazylibrarian-ffmpeg` branch in the repo. :)

PR #38, opened 2020-05-09
lazylibrarian-ffmpeg: initial release

mKeRix:
Follow up to #37, the content of the mod is the same as in my repo. It's quite simple and just based on the other examples on here. I've tested my published version of the mod with my personal install and it is working fine.


root/etc/cont-init.d/98-ffmpeg
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/with-contenv bash
+
+echo "**** installing ffmpeg ****"
+apt update && apt install -y ffmpeg
```
nemchik, line 4: Is the apt update here needed? Does it make a difference?

mKeRix: The lazylibrarian image ships without the package lists afaik, so without ffmpeg won't be found. I oriented myself on the other add-ons and only ran the update - if you want I can also do the proper "Docker-style" package install and clean up the package lists again afterwards.

nemchik: The way you have it is fine. Just wanted to check. You are correct the image is cleaned up. The mod should be fine like this though.

CHBMB: @nemchik Yeah, whenever you want to install a package in a Ubuntu container you have to `apt update` first, not the case with Alpine and `apk` though.

aptalca: it should really be `apt-get update` and `apt-get install` so it doesn't complain about tty

also, @mKeRix please add the maintainer label to the dockerfile and list yourself like here: https://github.com/linuxserver/docker-mods/blob/code-server-nodejs/Dockerfile#L3

mKeRix: @aptalca done!

PR #39, opened 2020-05-20
Fix PATH grep to catch export PATH as well as PATH

n-i-x:
When this mod adds to the PATH it adds it and exports it. The original grep doesn't catch that so it gets added again on container restart. This PR fixes that.

This will need to also get merged into the new -multi branch.


aptalca:
merged in #40

PR #45, opened 2020-05-25
code-server-powershell: fix tarball naming, switch to jq

JamiePhonic:
Powershell Tarballs dont use underscores in their file names, unlike the distro packages which do.
For example:
Linux x64 TAR: https://github.com/PowerShell/PowerShell/releases/download/v7.0.1/powershell-7.0.1-linux-x64.tar.gz
Debian 10 DEB: https://github.com/PowerShell/PowerShell/releases/download/v7.0.1/powershell_7.0.1-1.debian.10_amd64.deb

EDIT (by maintainer): also switch to jq for initial parsing of api result. With long releases, the github api response formatting is not consistent and awk failed to retrieve the version tag about 50% of the time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,10 @@ env:
     - BASEIMAGE="code-server"
     - MODNAME="powershell"
 
+before_install:
+  - sudo apt-get update
+  - sudo apt-get -y install jq
+
 jobs:
   include:
     - stage: PR-BuildImage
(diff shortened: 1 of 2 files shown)


aptalca:
@JamiePhonic thanks for the PR, I should have noticed the pushed image size of 463 bytes 😅

JamiePhonic:
@aptalca Don't worry about it, these things happen :P
I only noticed becasue i'm activley developing some scripts in powershell and figured hosting it Code-Server makes it easier to work with since no matter what device i use, it'll always be the latest version of the code and everything will be setup the way i like it :D

aptalca:
I absolutely agree. I do 95% of my dev work in code server now. Being able to jump on any computer anywhere in the world and access my dev environment is awesome.

PR #47, opened 2020-05-29
Add /config/go/bin to PATH for user binaries

n-i-x:
This change adds user installed go binaries to the path

### DIFF
--- a/root/etc/cont-init.d/98-golang
+++ b/root/etc/cont-init.d/98-golang
@@ -7,7 +7,6 @@ if grep -q -E '^(export )?PATH=' /etc/services.d/code-server/run; then
   elif ! grep -q -E '^(export )?PATH=.*/config/go/bin:.*' /etc/services.d/code-server/run; then
     sed -i 's/PATH=/PATH=\/config\/go\/bin:/g' /etc/services.d/code-server/run
   fi
-  fi
 else
   sed -i '/^#!\/usr\/bin/a \\n# Added by codeserver-golang\nexport PATH=/config/go/bin:$PATH:/usr/local/go/bin' /etc/services.d/code-server/run
 fi
--- a/root/etc/cont-init.d/98-golang
+++ b/root/etc/cont-init.d/98-golang
@@ -4,7 +4,10 @@ echo "ensuring golang is in PATH"
 if grep -q -E '^(export )?PATH=' /etc/services.d/code-server/run; then
   if ! grep -q -E '^(export )?PATH=.*/usr/local/go/bin.*' /etc/services.d/code-server/run; then
     sed -i '/PATH/ s/$/:\/usr\/local\/go\/bin/' /etc/services.d/code-server/run
+  elif ! grep -q -E '^(export )?PATH=.*/config/go/bin:.*' /etc/services.d/code-server/run; then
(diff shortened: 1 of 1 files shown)


root/etc/cont-init.d/98-golang
```diff
@@ -4,7 +4,10 @@ echo "ensuring golang is in PATH"
 if grep -q -E '^(export )?PATH=' /etc/services.d/code-server/run; then
   if ! grep -q -E '^(export )?PATH=.*/usr/local/go/bin.*' /etc/services.d/code-server/run; then
     sed -i '/PATH/ s/$/:\/usr\/local\/go\/bin/' /etc/services.d/code-server/run
+  elif ! grep -q -E '^(export )?PATH=.*/config/go/bin:.*' /etc/services.d/code-server/run; then
+    sed -i 's/PATH=/PATH=\/config\/go\/bin:/g' /etc/services.d/code-server/run
+  fi
```
aptalca, line 9: this `fi` seems extra

root/etc/cont-init.d/98-golang
```diff
@@ -4,7 +4,10 @@ echo "ensuring golang is in PATH"
...
   fi
 else
-  sed -i '/^#!\/usr\/bin/a \\n# Added by codeserver-golang\nexport PATH=$PATH:/usr/local/go/bin' /etc/services.d/code-server/run
+  sed -i '/^#!\/usr\/bin/a \\n# Added by codeserver-golang\nexport PATH=/config/go/bin:$PATH:/usr/local/go/bin' /etc/services.d/code-server/run
```
aptalca, line 12: Can you move `config/go/bin` to the end, next to `/usr/local/go/bin`? It's triggering my OCD :)

root/etc/cont-init.d/98-golang
```diff
@@ -4,7 +4,10 @@ echo "ensuring golang is in PATH"
 if grep -q -E '^(export )?PATH=' /etc/services.d/code-server/run; then
   if ! grep -q -E '^(export )?PATH=.*/usr/local/go/bin.*' /etc/services.d/code-server/run; then
     sed -i '/PATH/ s/$/:\/usr\/local\/go\/bin/' /etc/services.d/code-server/run
+  elif ! grep -q -E '^(export )?PATH=.*/config/go/bin:.*' /etc/services.d/code-server/run; then
+    sed -i 's/PATH=/PATH=\/config\/go\/bin:/g' /etc/services.d/code-server/run
```
aptalca, line 8: why not use the same sed method for both?

root/etc/cont-init.d/98-golang
n-i-x: I was going to do that, but then considered the case where a user wanted their binaries first in the path. This does that.

root/etc/cont-init.d/98-golang
n-i-x: Because the first appends the PATH to the end and the second puts it first. See my previous reply.

root/etc/cont-init.d/98-golang
n-i-x: not sure what's wrong with me today...

root/etc/cont-init.d/98-golang
aptalca: ok, that makes sense

PR #60, opened 2020-08-17
mariadb-mysqldump-cron

nemchik:
This mod adds cron schedules that run mysqldump on mariadb. The default will create daily backups of all databases with 7 day rotation and weekly backups with 4 week rotation. You can edit `/config/crontabs/mysqldump` to change the schedules, but the file cannot be renamed.


nemchik:
Converted to draft to prevent accidental merging while working with GHA/GHCR

nemchik:
This can be accomplished with the universal cron mod. the helper script simply passes the root password environment variable. Closing.

PR #66, opened 2020-09-23
swag:geoip2influx adds geoip2 python script

gilbN:
This mod adds a python script that sends geo location metrics from the nginx access log to InfluxDB.
The mod will parse the access log for IPs and and convert them into geo metrics for InfluxDB. It will also send log metrics if enabled.


root/etc/cont-init.d/69-geoip2influx
```diff
@@ -0,0 +1,56 @@
...
+    echo '-------------------------------------'
+    mkdir  /config/geoip2db
+fi
+
+# Fetch latest or update existing
+if [[ ! -d geoip2influx/.git ]]; then
+  echo '---------------------------'
+  echo '| Installing Geoip2Influx |'
+  echo '---------------------------'
+  git clone --progress --verbose https://github.com/gilbN/geoip2influx.git
```
aptalca, line 27: Does this have to reside in the config folder? Wouldn't it be better to ship this inside the mod image (tarball)? We can set a jenkins trigger to build and push the mod whenever there is an upstream release or commit

gilbN: Sounds like a good idea!

gilbN: The python script creates the `geoip2influx.log` file in the same directory as itself, so we need to touch and symlink that to /log/geoip2influx/ for example.

aptalca (approved):

PR #71, opened 2020-10-05
Add universal cron

nemchik:
Tested: `nemchik/mods:universal-cron`


drizuid (approved):
i'm assuming crond is part of alpine's busybox since there's no install check for alpine in 95-cron

nemchik:
> i'm assuming crond is part of alpine's busybox since there's no install check for alpine in 95-cron

Correct

PR #87, opened 2020-10-23
universal-docker: initial release

aptalca:


root/etc/cont-init.d/98-docker
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/with-contenv bash
```
nemchik, line 1: Filename 95 for universal

aptalca: 🤦‍♂️

aptalca: force pushed

nemchik (approved):

PR #88, opened 2020-10-25
Replace Travis with GHA

nemchik:
Replaces Travis with GitHub Actions workflow.

Requires we set secrets in the repo setting https://github.com/linuxserver/docker-mods/settings/secrets for DOCKERPASS and DOCKERUSER (same as how it's setup in Travis).

Users only need to edit the 2 environment variables

```yaml
      BASEIMAGE: "baseimagename" #replace
      MODNAME: "modname" #replace
```

Image push only happens if the branch matches the variables (so branches in development would not push). Build happens on all branches (so we can confirm build is successful without pushing images).

For now, new images should use this. Long term goal would be to replace Travis with GHA on all existing branches.


nemchik:
P.S. Confirmed with https://github.com/linuxserver/docker-mods/pull/89 that PRs will build but not push.

nemchik:
Also confirmed pushing works https://github.com/nemchik/docker-mods/runs/1303687783

aptalca:
This looks great. We should probably also update the readme on the master branch with gha instead of travis after this is merged

aptalca:
`DOCKERUSER` and `DOCKERPASS` secrets are now added to repo

aptalca (approved):

aptalca (approved):

PR #91, opened 2020-10-25
Add Swag-cloudflare-real-ip

Roxedus:
Adds a script that on execution (start of the container) creates/modifies a file with Cloudflares IPs for X-Forward-For


Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM scratch
 
-LABEL maintainer="username"
+LABEL maintaner="Roxedus"
```
aptalca, line 3: Typo in `maintainer`

root/etc/cont-init.d/98-cloudflare
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/with-contenv bash
```
aptalca, line 1: Can you rename this file to be the same as the mod name, `98-cloudflare-real-ip` ?

Roxedus: Done

aptalca (approved):

PR #131, opened 2020-11-02
Switch to GHA, add GHCR

nemchik:


aptalca:
this is unpublished as of yet

PR #138, opened 2020-11-03
Add information about GHCR

nemchik:


Dockerfile.complex
```diff
@@ -1,5 +1,5 @@
 ## Buildstage ##
-FROM lsiobase/alpine:3.12 as buildstage
+FROM ghcr.io/linuxserver/baseimage-alpine::3.12 as buildstage
```
aptalca, line 2: extra colon here

nemchik: Good catch. Weak copy pasta on my part.

aptalca (approved):

PR #140, opened 2020-11-07
Mariadb: mariadb backup adds mariadb-backup to the container

drizuid:
closes linuxserver/docker-mariadb#56


nemchik (changes requested):
We no longer use the Travis file.

Issue #164, opened 2021-01-21
[Swag-cloudflare-realip] Mod kills container if it cannot access Cloudflare


rg9400:
This mod accesses cloudflare's IP page and places them into a file. However, in my case, sometimes my DNS server is not up by the time the Swag container spins up (even if I set Swag to depend on my DNS container). Since this mod runs right away on startup, there is not a huge gap to allow the DNS container to fully spin up.

If it fails to access the website due to DNS resolution errors, it does not actually list out any IPs, and the created file creates a fatal error in nginx -- `nginx: [emerg] invalid number of arguments in "set_real_ip_from" directive in /config/nginx/cf_real-ip.conf:1`.

Ideally, the mod should be able to retry access to the CF page if inaccessible, and maybe if it still cannot access it after a sane number of retries, it has a proper fallback that does not brick the entire container.

Roxedus:
fixable with https://github.com/linuxserver/docker-mods/pull/169

PR #167, opened 2021-01-22
[Swag-cloudflare-realip] Added internet accessibility pre-check and timeouts

N3rdP1um23:
### Edits
- Perform a pre-check to make sure the internet (Cloudflare specifically) is accessible
- Added a reasonable retry count of 25 before the script will fail out
- Added connection timeouts to help with smoother curl failures

### Reasoning
After talking with https://github.com/rg9400, in relation to his issue https://github.com/linuxserver/docker-mods/issues/164, I've added an internet accessibility pre-check allowing up to 25 retries before it fails out. Also, I've added a connection timeout to the curl commands to make the failure rate smoother after 5s if the connection to Cloudflare couldn't be made.

### Use Case
In RG's case, even though he has it set up where Swag will execute after AdGuardHome is active, AGH still has an initialization process and causing this docker-mod to fail as there isn't an active internet connection at the time.

### DIFF
--- a/root/etc/cont-init.d/98-cloudflare-real-ip
+++ b/root/etc/cont-init.d/98-cloudflare-real-ip
@@ -2,9 +2,28 @@
 # shellcheck shell=bash
 # shellcheck disable=SC2046
 
+internet_access=0
+retries_left=25
+
+while [[ ${internet_access} -eq 0 ]]; do
+	curl --connect-timeout 5 -I https://www.cloudflare.com/ips-v4 2> /dev/null
+	
+	rc=$?
+	
+	if [[ ${rc} -eq 0 ]]; then
+		internet_access=1
+	fi
+
+	if [[ ${retries_left} -eq 0 ]]; then
+		internet_access=2
+	fi
+
+	retries_left=$(( ${retries_left} - 1 ))
+done
+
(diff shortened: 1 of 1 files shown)


N3rdP1um23:
Hey! I've cleaned up my previous edits to be leaner overall and fall back if there's no internet access and a cf_real-ip.conf file already exists. If there's no internet and the file doesn't already exist, then it'll create it with the real ip from the docker interface.

Roxedus:
I am still not sure this is something that should be checked for in the mod exclusively, as there are other init processes in SWAG that also potentially needs internet access, such as https://github.com/linuxserver/docker-swag/blob/master/root/etc/cont-init.d/50-config

N3rdP1um23:
Oh no worries at all and yeah that's understandable. If you'd like, I could submit a PR for the other scripts also or close off this PR, either one is good with me 🙂

Roxedus:
I'm actually thinking a universal mod that pauses the whole init process when there is no internet

N3rdP1um23:
Yeah, that's a much better idea. Let me know if there's anything that I can help with

Roxedus:
I created the branch https://github.com/linuxserver/docker-mods/tree/universal-wait-for-internet, you can PR a loop that waits for internet into that one.

PR #169, opened 2021-02-05
[universal-wait-for-internet] Upgraded repo with actual wait-for-internet commands and information

N3rdP1um23:
Here's the PR for an earlier chat I had with Roxedus (https://github.com/linuxserver/docker-mods/pull/167). This mod will allow any docker to wait for an active internet connection before being further executed. This can be used in a case where there are pre-network requirements that need to be executed before a container has an active internet connection. As an example, waiting for an AdGuardHome docker container to finish initializing before later containers can have an active connection.


.github/workflows/BuildImage.yml
```diff
@@ -4,8 +4,8 @@ on: [push, pull_request, workflow_dispatch]
 
 env:
   ENDPOINT: "linuxserver/mods" #don't modify
-  BASEIMAGE: "replace_baseimage" #replace
-  MODNAME: "replace_modname" #replace
+  BASEIMAGE: "universal" #replace
+  MODNAME: "wait_for_internet" #replace
```
aptalca, line 8: these need to be dashes

N3rdP1um23: Whoops, my bad. All good now!

Roxedus:
This should probably run at the very start of the container, delaying all other startup scripts. I suppose 15-wait-for-internet should do it.

N3rdP1um23:
> This should probably run at the very start of the container, delaying all other startup scripts. I suppose 15-wait-for-internet should do it.

Updated numbering as suggested!

BrodyStone21:
Sorry if this is a beginner question, but I'm up to date and have been having this issue for months. Could someone tell me how I can go about fixing it?

aptalca:
What is the issue? This is a PR that was merged, not an issue report.

BrodyStone21:
> What is the issue? This is a PR that was merged, not an issue report.

If SWAG starts before Pihole, I have to `nano ~/.config/swag/nginx/cf_real-ip.conf` and delete the lines that would only have a semi colon. Then, I have to restart the swag container `docker restart swag` and everything will work. Please note, I _think_ this is because of Pihole, but I could be wrong. It usually happens after rebooting my server, so that's what I assume. 

I've already cleared my logs today, but I'll try and give an example off the top of my head

```cf_real-ip.conf
;
;
123.456.789 ;
987.654.321 ;
207.123.444 ;
```

Notice how there are two lines with no ip? If I run `docker logs swag` I'll get the error `nginx: [emerg] invalid number of arguments in "set_real_ip_from" directive in /config/nginx/cf_real-ip.conf:1`

I'm here because #164 referenced here and said that this fixed it.

PR #186, opened 2021-03-09
swag-auto-reload: initial release

aptalca:


root/etc/services.d/inotify/run
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/with-contenv bash
+
+DEFAULT_WATCH=( \
+  /config/nginx/authelia-location.conf \
+  /config/nginx/authelia-server.conf \
+  /config/nginx/geoip2.conf \
+  /config/nginx/ldap.conf \
+  /config/nginx/nginx.conf \
+  /config/nginx/proxy-confs \
```
Roxedus, line 9: Can we wildcard these? no need to trigger a reload on files that wont be reloaded

aptalca: inotify only loads a static list of files to watch and won't expand the list unless it's restarted, so if a new proxy conf file is enabled that wasn't previously in the list, it won't be added to the watch list. But if we watch a folder, it will pick up newly created files in that folder.

Roxedus (approved):

PR #189, opened 2021-03-11
Lidarr: flac2mp3 release 1.2.1

TheCaptain989:
- Added Variable Bit Rate option
- Fixed cover art conversion issue
- Fixed file permission/owner issue
- Reworked some bash logic
- Updated documentation


.github/workflows/BuildImage.yml
```diff
@@ -15,8 +15,15 @@ jobs:
 
       - name: Build image
         run: |
-          docker build --no-cache -t ${{ github.sha }} .
-
+          # Build variables
+          VERSION=$(git describe --tags --always)
```
Roxedus, line 19: This wont return anything useful

TheCaptain989: It'll have the first 5 chars of the commit string.
I have this show up in the log on container start, mainly so I can keep track of future bug reports.  But I'm totally open to a better way to tag my builds.

Roxedus: I would just use the github.sha, and truncate it, if you dont need the whole string. ie `${GITHUB_SHA:0:5}`

TheCaptain989: Got it.  Updated as requested.
I'll work on a better, easier to maintain human-readable versioning scheme later.

aptalca (approved):

PR #193, opened 2021-03-22
Swag geoip2influx

gilbN:
Added influxdb version note.


README.md
```diff
@@ -40,6 +40,8 @@ Get your licence key here: https://www.maxmind.com/en/geolite2/signup
 
 ### InfluxDB 
 
+#### InfluxDB v2.x is not supported. Use v1.8.x.
```
Roxedus, line 43: Maybe add a tag example on how to do this?

gilbN: It's influxdb agnostic, so I don't think it needs an example for just one specific installation method.

PR #201, opened 2021-04-13
update filter as they release zips too now

aptalca:


aptalca:
the build will fail until MS updates the json release files for 3.x and 2.x

aptalca:
Still should be safe to merge since the current build is also failing

Roxedus (approved):

PR #204, opened 2021-05-25
universal: cloudflared mod initial release

Spunkie:
Docker mod to build and install [`cloudflared`](https://github.com/cloudflare/cloudflared) then create/configure/route/enable a cloudflare tunnel via `cloudflared` and the cloudflare v4 API during startup.

The init script has been tested on x86-64 and arm64 but I don't have anything to test armhf on atm. I've done a bunch of spot testing and testing as a standalone script but it hasn't yet been e2e tested as an actual "docker mod".


aptalca:
@Spunkie I added versioning to the image via tags

One last issue is, there is no systemd or rc.d in our images, so you'll have to create an s6 service for cloudflared

aptalca (approved):

PR #209, opened 2021-06-03
swag-auto-proxy: add label `swag_address`

aptalca:


root/app/auto-proxy.sh
```diff
@@ -100,7 +104,11 @@ DUDE
             done
             echo "}" >> "/etc/nginx/http.d/auto-proxy-${CONTAINER}.subdomain.conf"
         fi
+        if [ -z "${swag_address}" ]; then
+            swag_address="${CONTAINER}"
+        fi
         sed -i "s|<container_name>|${CONTAINER}|g" "/etc/nginx/http.d/auto-proxy-${CONTAINER}.subdomain.conf"
```
TheSpad, line 110: Shouldn't this be using swag_address instead of CONTAINER?

aptalca: D'oh
Nice catch

Roxedus (approved):

PR #234, opened 2021-08-06
Code server java11

EGOISTK21:
Add java11 mod for code-server


aptalca (approved):

binarymelon:
@aptalca @EGOISTK21 Could this be universal?

binarymelon:
Easily add java to other containers.  Is there any reason why this wouldn't work?  I applied it to the calibre container and it seems to work fine.

binarymelon:
For calibre specifically, it's because there is a plugin that has a java dependency.

https://github.com/w3c/epubcheck

aptalca:
Gotcha, thanks. Then yes, it's safe to use this mod on calibre (but only the `latest` tag. It won't work on the `arch` tag as it is archlinux based).

PR #257, opened 2021-11-14
ProjectSend: translations

alexschomb:
ProjectSend now provides [translations on their website](https://www.projectsend.org/translations/) as zip file. Unfortunately, I couldn't find an easy solution for importing the contents of those zip files into the appropriate directories. (volume mounts disturbed the build process of the image) So I created a Docker Mod to automatically install languages from the website by given ``TRANSLATIONS`` environment variable. This variable can be a comma separated list of language short codes (as provided on the translations' website).

Please note: This is my first Docker Mod. Please let me know in case I did any mistakes.

PR #264, opened 2021-11-30
Projectsend translations

alexschomb:
Please see https://github.com/linuxserver/docker-mods/pull/257

I hope I'm doing this right.


Dockerfile
```diff
@@ -1,6 +1,10 @@
...
 
-LABEL maintainer="username"
+LABEL maintainer="alexschomb"
 
 # copy local files
 COPY root/ /
+
+# install translations
+RUN chmod +x /install-translations.sh
+CMD /install-translations.sh
```
aptalca, line 10: these two lines won't work here as it needs to be a single layer image, so only the `COPY` line is allowed in the final image

root/install-translations.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -e
```
aptalca, line 1: this file needs to be an init script dropped into `root/etc/cont-init.d` with the appropriate filename/number like here: https://github.com/linuxserver/docker-mods/blob/code-server-docker/root/etc/cont-init.d/98-docker

alexschomb (commented):
Thanks for your feedback. I did update the files to reflect the requirements.

aptalca:
My apologies. I don't how I let this one sit for so long. I'll clean this up and merge.

aptalca (approved):

alexschomb:
@aptalca Sweet, thanks for the merge!

PR #273, opened 2021-12-14
Code server svn

### LINKED ISSUE #272: code-server MOD: SVN
I've created a mod for code-server that installs Subversion and the SVN extension in code-server: [code-server-svn](https://github.com/RoboMagus/docker-mods/tree/code-server-svn).

Please create a ```code-server-svn``` branch so that I can create a PR to submit this mod.

RoboMagus:
A docker mod that installs Subversion and SVN extension into code-server
Closes #272


aptalca:
The two init files can be combined into one as 98-subversion, no need to have two separate ones
Also, the code-server mods are being made available for openvscode-server as well, can you make the extension install step compatible with both with an if perhaps?

RoboMagus:
Thanks for the feedback. 
The two init files have been merged, However I currently have no experience with the openvscode-server and was unaware of it's compatibility with code-server mods. I'd have to look into the changes required to get the extension installed for that as well (correct user, directory, etc.). 

That will be an issue to be tackled later when time once more permits. Then I can also update ```code-server-extension-arguments```.

Roxedus:
When you revisit this, also move the apt-update into its own file, called `95-apt-get`, like this https://github.com/linuxserver/docker-mods/blob/code-server-php/root/etc/cont-init.d/95-apt-get

RoboMagus:
@Roxedus
I had just merged all in it files based on a comment things could be combined. Is there a valid reason this should all be fixed?

If so this justification should (more visibly) be included in the documentation.

Roxedus:
We don't want to unnecessarily slow down the startup, as each mod used to to apt update, it could takes a while when you have a lot of mods

Roxedus:
@RoboMagus I have abstracted this away, you just need to use the helper. Like I have done here https://github.com/linuxserver/docker-mods/blob/code-server-terraform/root/etc/cont-init.d/98-terraform#L8

I have also slightly modified the template branch to mention the 95-apt-get

RoboMagus:
@Roxedus 
Thanks for the example! I did not know of the existence of that abstraction. That's a big helper for sure.

RoboMagus:
@Roxedus Could you please take another look for approval?

Also curious about your view on my previous comment: 
>  Looking at the implementation of `install-extension` I'm wondering if `--user-data-dir /config/data` needs to be added? In the examples I can find where extensions are pre-installed, this flag seems to be always provided.

Thanks :)

Roxedus:
There is no user involved in the extension plugin process

PR #292, opened 2022-01-14
Added more info to f2b

quietsy:
Added the last banned IP and hovering over it shows the data relating to the ban


aptalca (approved):

root/dashboard/swag-f2b.py
```diff
@@ -4,7 +4,22 @@
 
 con = sqlite3.connect("/config/fail2ban/fail2ban.sqlite3")
 cur = con.cursor()
-results = cur.execute("SELECT jails.name, COUNT(bans.ip) AS bans FROM jails LEFT JOIN bans ON jails.name=bans.jail GROUP BY jails.name").fetchall()
+results = cur.execute("""
+    SELECT jails.name, 
+    COUNT(bans.ip) AS bans,
```
GuiPoM, line 9: @quietsy and @aptalca I was having a look to this PR as I was confused by what the dashboard was showing me: 8 bans, when there is none technically speaking:
image
I now better understand that bans is a count on table entries, but these entries may be unbanned at the time of this check, which was my case.
Dashboard vs SQLDBLite:
image

I wanted to submit an issue but I am not so sure if this is here a bug or a feature. I would find "Active Bans" (timeofban + bantime > now) more useful for users than "Bans". Maybe both could be useful, I honestly don't know, and as you seem to be both involved in this topic more than I am, I would like your opinion on it.

Many thanks for your contributions on the dashboard mod, it changed the way I am monitoring my swag container and helps me to focus my attention on important information I may have miss otherwise 👍

PR #293, opened 2022-01-14
Build goaccess from source

quietsy:


root/etc/cont-init.d/98-dashboard-config
```diff
@@ -2,12 +2,17 @@
 
 echo "Applying the SWAG dashboard mod..."
 
-apk add --no-cache --upgrade goaccess
-
+apk add libmaxminddb
```
aptalca, line 5: this should have `--no-cache`, which is the rough equivalent of `apt update`, but then again, the package is already installed in swag

Dockerfile
```diff
@@ -1,6 +1,33 @@
...
+    glib-dev \
+    libmaxminddb-dev \
+    ncurses-dev && \
+  mkdir -p /build && \  
+  mkdir -p /goaccess && \  
+  echo "**** build goaccess ****" && \  
+  git clone --shallow-submodules --recurse-submodules https://github.com/allinurl/goaccess.git /goaccess && cd /goaccess && \
+  autoreconf -fiv && \
+  ./configure --enable-utf8 --enable-geoip=mmdb && \
+  make DESTDIR="/build" install && \
```
aptalca, line 19: this would build it for amd64 only, no?

aptalca (approved):

PR #295, opened 2022-01-15
Swag dashboard - use fontawesome icons

GilbN:
* Added Font Awesome icons instead of the spans with green-circle/red-circle class on the `Available` and `Proxied` columns
* Added an info icon on the f2b table as it wasn't too obvious that you could hover over the last banned ip

@quietsy This is my first time doing any php, but it looks to be working.


quietsy (approved):

quietsy:
Looks great!

aptalca (approved):

PR #311, opened 2022-02-05
Add nvm mod for code-server

xiaohai-huang:
A docker mod that installs nvm into code-server


root/etc/cont-init.d/98-nvm
```diff
@@ -0,0 +1,24 @@
...
+[ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"
+
+echo "**** installing the current stable LTS release of Node.js ****"
+nvm install --lts
+
+echo "**** switching to the current stable LTS release of Node.js  ****"
+nvm use --lts
+
+echo "**** installing yarn  ****"
+npm install -g yarn
```
aptalca, line 20: Do all these operations need to run on every container start? Can you make some or all of them conditional?

fabricionaweb: You right, it dont need.

aptalca: not looking so good for this mod since the `maintainer` has not responded in months

PR #321, opened 2022-02-11
code-server:julia initial commit

MateoPeri:


root/etc/cont-init.d/98-julia
```diff
@@ -0,0 +1,51 @@
...
+gh_lates() {
+  curl -s https://api.github.com/repos/$1/$2/releases/latest \
+    | grep "tag_name" \
+    | cut -d : -f 2,3 \
+    | tr -d '\"\ v,'
+}
+
+download_vsix() {
+  wget -qO $1.$2-$3.vsix https://$1.gallery.vsassets.io/_apis/public/gallery/publisher/$1/extension/$2/$3/assetbyname/Microsoft.VisualStudio.Services.VSIXPackage
+}
```
Roxedus, line 12: Please use the helper function to install extentions. https://github.com/linuxserver/docker-mods/tree/template#tips-and-tricks

root/etc/cont-init.d/98-julia
```diff
@@ -0,0 +1,51 @@
...
+
+# replace these with specific versions if needed (i.e 1.7.2 for Julia)
+JULIA_VERSION=$(gh_latest julialang julia)
+JULIA_VSCODE_VERSION=$(gh_latest julia-vscode julia-vscode)
+VSCODE_FORMAT_VERSION=1.0.4
+JULIA_FORMATTER_VERSION=$(gh_latest singularitti vscode-julia-formatter)
+ANACONDA_VERSION=latest
+
+# Julia
+cd /usr/local/bin \
```
Roxedus, line 22: I would rather move the binary into bin after unpack, if doable.

root/etc/cont-init.d/98-julia
```diff
@@ -0,0 +1,51 @@
...
+  && sudo mkdir julia_dir \
+  && cd julia_dir \
+  && sudo wget -q -q https://julialang-s3.julialang.org/bin/linux/x64/`echo ${JULIA_VERSION} | cut -d. -f 1,2`/julia-${JULIA_VERSION}-linux-x86_64.tar.gz \
+  && sudo tar fxz julia-${JULIA_VERSION}-linux-x86_64.tar.gz \
+  && sudo rm -R julia-${JULIA_VERSION}-linux-x86_64.tar.gz \
+  && cd .. \
+  && sudo ln -s julia_dir/julia-${JULIA_VERSION}/bin/julia julia
```
Roxedus, line 29: No need for sudo (or &&), as init script runs as root.

root/etc/cont-init.d/98-julia
```diff
@@ -0,0 +1,51 @@
...
+  # vscode-format-context-menu
+  && download_vsix lacroixdavid1 vscode-format-context-menu ${VSCODE_FORMAT_VERSION} \
+  && code-server --install-extension lacroixdavid1.vscode-format-context-menu-${VSCODE_FORMAT_VERSION}.vsix \
+  && sudo rm lacroixdavid1.vscode-format-context-menu-${VSCODE_FORMAT_VERSION}.vsix \
+  # vscode-julia-formatter
+  && download_vsix singularitti vscode-julia-formatter ${JULIA_FORMATTER_VERSION} \
+  && code-server --install-extension singularitti.vscode-julia-formatter-${JULIA_FORMATTER_VERSION}.vsix \
+  && sudo rm singularitti.vscode-julia-formatter-${JULIA_FORMATTER_VERSION}.vsix
+  # better-toml
+  #&& code-server --install-extension bungcip.better-toml
```
Roxedus, line 47: This is not valid shellcode.

root/etc/cont-init.d/98-julia
```diff
@@ -0,0 +1,51 @@
...
+  && sudo wget -q -q https://julialang-s3.julialang.org/bin/linux/x64/`echo ${JULIA_VERSION} | cut -d. -f 1,2`/julia-${JULIA_VERSION}-linux-x86_64.tar.gz \
+  && sudo tar fxz julia-${JULIA_VERSION}-linux-x86_64.tar.gz \
+  && sudo rm -R julia-${JULIA_VERSION}-linux-x86_64.tar.gz \
+  && cd .. \
+  && sudo ln -s julia_dir/julia-${JULIA_VERSION}/bin/julia julia
+
+
+RUN cd /home/coder/project/ \
+  && sudo wget -q https://github.com/julia-vscode/julia-vscode/releases/download/v${JULIA_VSCODE_VERSION}/language-julia-${JULIA_VSCODE_VERSION}.vsix \
+  && code-server --install-extension language-julia-${JULIA_VSCODE_VERSION}.vsix \
```
Roxedus, line 34: We want to keep code-server mods compatible with openvscode

Roxedus (changes requested):

root/etc/cont-init.d/98-julia
MateoPeri: Is there documentation on how to use this tool?

root/etc/cont-init.d/98-julia
MateoPeri: Changed this to account for both errors.

root/etc/cont-init.d/98-julia
MateoPeri: Sorry, I did not notice this, it was originally part of a Dockerfile.

root/etc/cont-init.d/98-julia
MateoPeri: What are the incompatibilities? I am not knowledgeable with the differences.

root/etc/cont-init.d/98-julia
Roxedus: We have a example here https://github.com/linuxserver/docker-mods/blob/code-server-terraform/root/etc/cont-init.d/98-terraform#L8

aptalca:
Also, I don't believe there is wget in there so you'll need to use curl instead

root/etc/cont-init.d/98-julia
MateoPeri: I am now using the helper instead of `wget`ing the .vsix files and installing them. In your example there was no version specified so I assume it just grabs the latest. It should also mean that the code is now `openvscode`-compatible.

root/etc/cont-init.d/98-julia
```diff
@@ -0,0 +1,33 @@
...
+}
+
+# replace these with specific versions if needed (i.e 1.7.2 for Julia)
+JULIA_VERSION=$(gh_latest julialang julia)
+
+# Julia
+curl -so julia.tar.gz https://julialang-s3.julialang.org/bin/linux/x64/`echo ${JULIA_VERSION} | cut -d. -f 1,2`/julia-${JULIA_VERSION}-linux-x86_64.tar.gz
+tar fxz julia.tar.gz
+rm -R julia.tar.gz
+mv julia-${JULIA_VERSION}/bin/julia /usr/local/bin
```
aptalca, line 17: this should be done in a tmp folder and cleaned up after extract and move

root/etc/cont-init.d/98-julia
```diff
@@ -0,0 +1,33 @@
...
+# vscode-format-context-menu
+install-extension lacroixdavid1.vscode-format-context-menu
+# vscode-julia-formatter
+install-extension singularitti.vscode-julia-formatter
+# better-toml
+install-extension bungcip.better-toml
+
+
+chown -R root:root /root
+chmod -R 600 /root/.ssh
```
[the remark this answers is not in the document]
MateoPeri: my bad, this used to be part of a custom dockerfile and these are leftovers. removed now.

root/etc/cont-init.d/98-julia
aptalca: alternatively, it would be nice to have some versioning enabled via either:
1. Check for updates and only download the new tarball if there is an update, or
2. Ship the tarball in the image, push updates to mod image when there is an update upstream (preferred, we can help with setting the update trigger on jenkins)

MateoPeri: Which directory would be best? I'm not that well versed with docker, would just creating a `julia_dir` folder in `/tmp` work?
For now I've gone with `/home/root`. Is that a bad practice?
The julia directory needs to persist for the program to work, but the `tar.gz` is removed.

aptalca: I don't see where you're telling the app where its folder is located. I only see moving the bin into the local bin folder. `/home/root` is definitely not the right place as code server is not running as root, and the user's home folder is not set to it (and it's not persistent).

MateoPeri: I replaced the move command with a symlink. The new command is `ln -s $(pwd)/julia-${JULIA_VERSION}/bin/julia /usr/local/bin/julia`. So `/usr/local/bin/julia` should point to wherever we install it.
So now the workflow is download -> extract -> remove tar -> symlink to `/usr/local/bin/julia`.

If `/home/root` is not a good place, where else could we store it? As I said, I am quite new to Docker development. What is the standard way of keeping program files?

aptalca: I'm looking into julia install instructions and also see another issue, the binary you're pulling is x86_64 only, which is not going to work on arm32v7 and arm64v8 arches for these images

aptalca:
3 of the extensions fail to install:
```
Extension 'julia-vscode.julia-vscode' not found.
Make sure you use the full extension ID, including the publisher, e.g.: ms-dotnettools.csharp
Failed Installing Extensions: julia-vscode.julia-vscode
Installing extensions...
Extension 'yellpika.latex-input' not found.
Make sure you use the full extension ID, including the publisher, e.g.: ms-dotnettools.csharp
Failed Installing Extensions: yellpika.latex-input
Installing extensions...
Extension 'lacroixdavid1.vscode-format-context-menu' not found.
Make sure you use the full extension ID, including the publisher, e.g.: ms-dotnettools.csharp
Failed Installing Extensions: lacroixdavid1.vscode-format-context-menu
```

MateoPeri:
That was the reason for downloading the vsix files and then installing manually. VSCode cannot find some extensions, for some reason. Installing from the file is fine. The id and publisher are fine. You can check this at the VSCode marketplace (https://marketplace.visualstudio.com/items?itemName=lacroixdavid1.vscode-format-context-menu). That same id and publisher are being used to download the vsix files, so they work only sometimes.

I do not know why this happens. I was hoping someone could shed some light on this bug.

Funnily enough, the example provided by the command does not work:
```
> code-server --install-extension ms-dotnettools.csharp
Installing extensions...
Extension 'ms-dotnettools.csharp' not found.
Make sure you use the full extension ID, including the publisher, e.g.: ms-dotnettools.csharp
Failed Installing Extensions: ms-dotnettools.csharp
```

aptalca (approved):

MateoPeri:
got this in the logs when trying to run the container. It starts fine but julia is not installed. It appears `tar` is failing, but there are no other log entries that are julia-mod-related.

```
code-server    | [mod-init] Attempting to run Docker Modification Logic
code-server    | [mod-init] Applying linuxserver/mods:code-server-julia:latest files to container
code-server    |
code-server    | gzip: stdin: not in gzip format
code-server    | tar: Child returned status 1
code-server    | tar: Error is not recoverable: exiting now
code-server    | [s6-init] making user provided files available at /var/run/s6/etc...exited 0.
code-server    | [s6-init] ensuring user provided files have correct perms...exited 0.
code-server    | [fix-attrs.d] applying ownership & permissions fixes...
## after this it just boots normally
```

aptalca:
try again, it should work now

MateoPeri:
Tried again and got the same result. Deleted all containers and images and re-pulled again, both the base image and the mod, but still got the same issue.

This is my compose just in case, although it's not probably a compose issue. I see the the container trying to set up the mod, but failing, so that means that the env variable for the mod is set correctly (I assume).
```yaml
---
version: "2"
services:
  code-server:
    image: linuxserver/code-server
    container_name: code-server
    restart: unless-stopped
    environment:
      - TZ=Europe/Madrid
      - PGID=1000
      - DEFAULT_WORKSPACE=/data
      - DOCKER_MODS=linuxserver/mods:code-server-julia:latest
    volumes:
      - ./config:/config
      - /home/szeth/dev:/data
    ports:
      - 8443:8443/tcp
```


Something to note is that, before any `tar` command is run, there are some `echo`s ("Installing/updating Julia" for example) that should appear in the logs, but they don't.

aptalca:
there is no `:latest` tag, the tag itself is `code-server-julia`

MateoPeri:
Thanks, that fixed it

PR #329, opened 2022-02-16
Universal package install

RoboMagus:
Some rebase and cleanup required to make it able to merge.

Simple mod that allows users to add packages to be installed at startup to environment variable `INSTALL_PACKAGES`. These will then be installed using Apt / Apk.


README.md
```diff
@@ -1,25 +1,13 @@
...
-## Tips and tricks
-
-* To decrease startup times when multiple mods are used, we have consolidated `apt-get update` down to one file. As seen in the [nodejs mod](https://github.com/linuxserver/docker-mods/tree/code-server-nodejs/root/etc/cont-init.d)
-* Some images has helpers built in, these images are currently:
-    * [Openvscode-server](https://github.com/linuxserver/docker-openvscode-server/pull/10/files)
-    * [Code-server](https://github.com/linuxserver/docker-code-server/pull/95)
+For example, to install `rsync`, `git` and `nginx` add the following lines to your docker compose service:
+```yaml
+- DOCKER_MODS=linuxserver/mods:universal-package-install
+- INSTALL_PACKAGES=rsync git nginx
```
aptalca, line 12: It's best to have the vars separated by a delimiter like a comma instead of space, because var entry is handled differently between compose and cli

root/etc/cont-init.d/97-install-packages
```diff
@@ -0,0 +1,17 @@
...
+
+# Exit if no installable packages are provided
+if [ -z ${INSTALL_PACKAGES+x} ]; then
+    echo "**** No packages to install ****"
+    exit 0
+fi
+
+# Ubuntu
+if [ -f /usr/bin/apt ]; then
+    apt install -y --no-install-recommends ${INSTALL_PACKAGES}
```
aptalca, line 11: should be `apt-get`

README.md
RoboMagus: I did think about this initially, but this way the env. var can be passed directly into the apk / apt command. This would allow for passing in further command flags without it being messed up by additional parsing that would have to be done.

Applying propper quotes should solve that though. I see that in my example I did not yet include that.

root/etc/cont-init.d/97-install-packages
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/with-contenv bash
```
aptalca, line 1: file name needs to be 98-blah

README.md
aptalca: > This would allow for passing in further command flags

It's not proper to pass in command flags that way. Please use a delimiter for the package list and you can create a separate var for command flags if need be.
We'd rather not rely on quotes due to them being handled very differently between cli and the various different entry types in yaml.

RoboMagus: Understandable. Reworked this to use the same delimiting as using multiple mods.

root/etc/cont-init.d/97-install-packages
RoboMagus: fixed

aptalca (approved):

aptalca (changes requested):
Please set the maintainer label in the Dockerfile

RoboMagus:
Appologies for the missing maintainer label. Could've sworn that was set during previous changes...
Oh well. It's corrected now.

aptalca (approved):

PR #343, opened 2022-02-28
host f2bdb Patch 1

audioscavenger:
as discussed with @quietsy , created host f2bdb Patch 1 and Patch 2 to also read the host's f2bdb only if it's mounted in swag at the default location. simple and efficient, no documentation as requested

### DIFF
--- a/root/dashboard/swag-f2b-expand.py
+++ b/root/dashboard/swag-f2b-expand.py
@@ -26,14 +26,19 @@ if not os.path.exists(MOD_DASHBOARD_F2B_HOST_DB):
 
 con = sqlite3.connect(MOD_DASHBOARD_F2B_HOST_DB)
 cur = con.cursor()
-results = cur.execute("SELECT DISTINCT timeofban, bancount, ip from bans where jail = '%s' ORDER BY timeofban DESC LIMIT %s" % (jail, MOD_DASHBOARD_F2B_MAX_LINES)).fetchall()
-# results = cur.execute("SELECT * from bans where jail='sshd' limit 10").fetchall()
+results = cur.execute("""
+SELECT DISTINCT timeofban, data, ip 
+FROM bans 
+WHERE jail = '%s' 
+ORDER BY timeofban DESC 
+LIMIT %s 
+""" % (jail, MOD_DASHBOARD_F2B_MAX_LINES)).fetchall()
 con.close()
 formatted_results = [{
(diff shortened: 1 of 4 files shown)


quietsy:
Please combine to a single PR

root/dashboard/swag-f2b.py
```diff
@@ -1,8 +1,12 @@
 import json
 import sqlite3
+import sys
 
+f2bdb = "/config/fail2ban/fail2ban.sqlite3"
```
quietsy, line 5: Please use an environment variable and combine the results of swag's f2b with the custom f2b, there's no need to display the same section twice

quietsy: (And it won't require any changes to the php)

audioscavenger:
voila all is good now. using docker env variables, includes the expand button.
image
compose variables and volume needed to activate it:

volumes:
  - /var/lib/fail2ban/fail2ban.sqlite3:/var/lib/fail2ban/fail2ban.sqlite3:ro
environment:
  - MOD_DASHBOARD_F2B_HOST_DB=/var/lib/fail2ban/fail2ban.sqlite3    # swag mounted path to the host's fail2ban database; default is /var/lib/fail2ban/fail2ban.sqlite3
  - MOD_DASHBOARD_F2B_MAX_LINES=30                                  # max IPs to show per jail; default is 30

root/dashboard/swag-f2b-host.py
```diff
@@ -0,0 +1,32 @@
+import json
```
quietsy, line 1: There's a lot of duplicate code both here and in the php, better change the current code to handle the host f2b

root/dashboard/swag-f2b-expand.py
```diff
@@ -0,0 +1,45 @@
...
+    MOD_DASHBOARD_F2B_MAX_LINES = os.environ['MOD_DASHBOARD_F2B_MAX_LINES']
+if "TZ" in os.environ:
+    TZ = os.environ['TZ']
+
+if not os.path.exists(MOD_DASHBOARD_F2B_HOST_DB):
+    exit()
+
+con = sqlite3.connect(MOD_DASHBOARD_F2B_HOST_DB)
+cur = con.cursor()
+results = cur.execute("""
```
quietsy, line 29: The banned IPs can be returned together with the jails in the initial load

quietsy:
Added the host f2b part under `https://github.com/linuxserver/docker-mods/pull/351`, I'll consider the UI changes separately, thanks for the effort!

PR #345, opened 2022-03-07
Fixed code-server-ssl readme

MateoPeri:


root/etc/cont-init.d/98-ssl-config
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/with-contenv bash
 
-echo "**** adding --cert and --cert-key parameters to code-server startup ****"
-
-sed -i 's/\/bin\/code-server \\/\/bin\/code-server \\\n            --cert ${SSL_CERT_PATH} \\\n            --cert-key ${SSL_KEY_PATH} \\/g' /etc/services.d/code-server/run
+if [ -f "/etc/services.d/openvscode-server/run" ];
```
MateoPeri, line 3: this is a naive way to check if we are using openvscode-server or the regular one.

aptalca: that should be fine. You have to hard code the binary name anyway. We'll have to keep this in mind in case the binary name changes (it did recently)

aptalca (approved):

PR #351, opened 2022-03-19
Added support for more configurations

quietsy:
- Added support for internal access using `<server-ip>:81`
- Added support for host f2b and logs
- Added support for `proxy-confs` subfolders


TheSpad (approved):
LGTM

PR #377, opened 2022-05-17
Update auto-proxy.sh

Emalton:
Add support for TLS 1.3 in the auto generated config files.


aptalca:
This won't help you with portainer

PR #386, opened 2022-06-03
Add 127.0.0.1 to cloudflare-real-ip, list of IPs

Sblop:
Needed when using cloudflare tunnels. 
Real IP is not used in swag and all IPs were reported as 127.0.0.1. 
Adding this, solved the issue.


Roxedus:
I am pretty sure this would negatively impact users not using cloudflared, I propose this PR should rather document this need as a manual change needed by the user.

Sblop:
You are probably right… I will update documentation instead.

PR #393, opened 2022-07-11
[universal-stdout-logs] Use tail -F to follow rotated log files

brycied00d:
I appreciate this mod for keeping an eye on nginx logs and such. The lsio/swag rotates those log files daily, and unfortunately that causes this mod to stop working. I propose using tail's `-F` (a shortcut for `--follow=name --retry`) instead of `-f` so that this mod will continue tailing the newly-rotated log file.

### DIFF
--- a/root/etc/services.d/stdout-logs/run
+++ b/root/etc/services.d/stdout-logs/run
@@ -2,5 +2,5 @@
 
 TAIL_LOGS=$(echo "$LOGS_TO_STDOUT" | sed 's#|# -f #g')
 
-echo "Executing: tail -f $TAIL_LOGS"
-tail -f $TAIL_LOGS
+echo "Executing: tail -F $TAIL_LOGS"
+tail -F $TAIL_LOGS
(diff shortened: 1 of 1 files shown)


brycied00d:
I'm not sure of the proper etiquette for the docker-mods project, but I'd like to be sure the mod's maintainer sees this PR this is a ping to @alex-phillips

aptalca:
need to modify this line as well: https://github.com/linuxserver/docker-mods/pull/393/files#diff-00a302dd6eb4c97d2ac473fd1b10c373c01facfb096da67d84511f94e77c8571R3

brycied00d:
Thanks @aptalca - Modify it in what way? That line is just splitting apart the passed variable string of log files, and nothing about that logic needs to change as best as I can tell.

aptalca:
Are you sure it's unnecessary? Keep in mind that this mod needs to work on both alpine and ubuntu and tail may have different versions and capabilities between the two (alpine uses busybox versions of a lot of built in commands rather than gnu)

brycied00d:
Appreciate the feedback @aptalca - I did test and confirm this on an Alpine-based image (linuxserver/swag to be exact) before committing. I've never known a tail implementation that required `-f` multiple times, but I'm happy to be proven wrong. I have known some tail implementations that don't support `-F` and so that was the first thing I'd checked.

PR #396, opened 2022-07-13
Corrected sample country codes

E-Bloomer:
UK is not an ISO 3166-2 code and as such won't work as intended.


E-Bloomer:
Yes, but it's not in ISO 3166 which is what the .conf file says it uses.
(Also FIPS 10-4 was apparently withdrawn in 2008 and replaced with GENC which is based on ISO 3166)

aptalca:
looks like up until may 31st, maxmind accepted fips: https://dev.maxmind.com/geoip/release-notes/2022#data-changes-to-geoip-legacy-web-services but now it's deprecated

PR #399, opened 2022-07-15
Remove sample spam

lukyjay:
The dashboard is hard to read due to all the .conf.sample files provided by the SWAG container. This small tweak will only detect .conf files (excluding .conf.sample) to reduce the spam.

Side-note: Removing the .sample files does not work. They're re-created on container recreation/update.

### DIFF
--- a/root/dashboard/swag-proxies.py
+++ b/root/dashboard/swag-proxies.py
@@ -10,7 +10,7 @@ import urllib3
 
 def find_apps():
     apps = {}
-    file_paths = glob.glob("/config/nginx/**/**", recursive=True)
+    file_paths = glob.glob("/config/nginx/**/**.conf", recursive=True)
     auto_confs = glob.glob("/etc/nginx/http.d/*", recursive=True)
     file_paths.extend(auto_confs)
     for file_path in file_paths:


aptalca:
Please post a screenshot. I'm not seeing any sample files listed or displayed anywhere on the dashboard

lukyjay:
> Please post a screenshot. I'm not seeing any sample files listed or displayed anywhere on the dashboard

Yea sure, see attached before and after 🙂

Be curious to know why you don't see the other files

Screenshot_20220715-174637
Screenshot_20220715-174123

aptalca:
It's meant to show you the apps that are running on your system but are not being proxied via SWAG. It's a feature.

lukyjay:
> It's meant to show you the apps that are running on your system but are not being proxied via SWAG. It's a feature.

It's taking up a lot of visual space. There's no way to remove them considering the container re-downloads them.

Would love a few other people's opinions to see what the community preference is on this 🙂

aptalca:
It's not the community who's making the decision on something like this. As is, this PR will not be merged as it will only make a current feature useless, which is not desired. If you have ideas on alternative methods that don't break features, we're listening.

lukyjay:
Per above comment from Aptalca

PR #428, opened 2022-08-10
switch to hybrid (code-server-scikit-learn)

aptalca:


root/etc/s6-overlay/s6-rc.d/init-mod-code-server-scikit-learn-add-package/run
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/with-contenv bash
+
+if ! pip list 2>&1 | grep -q "scikit-learn"; then
```
thespad, line 3: This fails if pip isn't installed, which will always be the case because even if you include the python3 mod it won't have installed yet at this point.

aptalca: fixed

aptalca:
locally tested

thespad (approved):

PR #435, opened 2022-08-12
deprecate healthchecks-apprise

aptalca:


thespad:
This mod can be killed, we install apprise as part of the image now https://github.com/linuxserver/docker-healthchecks/blob/master/Dockerfile#L48

thespad (approved):

PR #441, opened 2022-08-12
switch to hybrid (lidarr-flac2mp3)

aptalca:
fyi @thecaptain989


TheCaptain989:
I'll be able to take a look at this in a couple days.

aptalca:
No worries, I just wanted to give you a heads up about making the mod compatible with s6v3, which is quite a bit different. This PR will make it compatible with both v2 and v3. Our baseimages are being updated to support v3 and the mods will all be made compatible before the downstream images are converted to v3. After that point, everything will be v3

TheCaptain989:
Okay, I think I'm caught up and I understand.
There's one tiny change that I would make to BuildImage.yml to properly support s6v3, but not having it doesn't break any functionality.

aptalca:
@TheCaptain989 I think I see what you're referring to. I just pushed an update in the last commit.

I also locally built this PR and pushed to `aptalca/mods:lidarr-flac2mp3`, if you could test that with the latest lidarr image and confirm it's working fine, we'll go ahead and merge this.

Thanks

TheCaptain989:
> @TheCaptain989 I think I see what you're referring to. I just pushed an update in the last commit.
> 
> I also locally built this PR and pushed to `aptalca/mods:lidarr-flac2mp3`, if you could test that with the latest lidarr image and confirm it's working fine, we'll go ahead and merge this.
> 
> Thanks

Cool!  I'll get started on testing.

TheCaptain989:
@aptalca, I finished my testing.
The mod and my script work fine!  The only thing I found is that my versioning did not come through, which I would expect since it was built locally.

aptalca:
oh yeah, the versioning is handled in the workflow and gets injected as a build arg (I didn't inject it when I built locally)

thespad (approved):

Issue #450, opened 2022-08-18
New branch request - jellyfin-rffmpeg


junkman690:
Request for new branch to be created to add rffmpeg support to Jellyfin https://github.com/junkman690/docker-mods/tree/jellyfin-rffmpeg

Allows remote transcoding on multiple hosts with optional WOL support.

aptalca:
Here you go: https://github.com/linuxserver/docker-mods/tree/jellyfin-rffmpeg

junkman690:
Thanks mate. Added pull request #451 which would close this one out

PR #451, opened 2022-08-18
jellyfin:rffmpeg initial release

junkman690:
Initial commit to add rffmpeg support to Jellyfin with optional WOL


junkman690:
This pull request will also close #450

aptalca:
Thanks for the PR. A couple of things. . .

Do you expect the `wol` script to be user customizable?

Do you expect `rffmpeg.yml` to be user customizable?

And do you expect either one to change frequently?

junkman690:
WOL was originally going to be user customisable, then I switched to docker ENV. I'll move it out of config. 

rffmpeg.yml is user customizable, however most users will be able to use as is without changing from defaults.

Both shouldn't change often upstream (bugs on my end aside for WOL). I can change the WOL script to pull only on initial once I move it out of config as it will then be upgraded on container recreation.

aptalca:
@junkman690 I just pushed a commit where I simplified the init quite a bit, precreated the symlinks and internalized the `wol` script. The rffmpeg script gets pulled from upstream master branch on every start. Not sure how often it will get updated, but that behavior can be changed later (perhaps when they start doing proper releases).

I pushed a local build of the latest changes to `aptalca/mods:jellyfin-rffmpeg`. Please test and let me know. Then I'll convert this to hybrid to support s6 v3 and then we can merge.

junkman690:
Hi mate, all the changes look good. My script did start getting a little messy. Did a quick test and can confirm it all worked correctly including the WOL.

junkman690:
Tested and can confirm rffmpeg and WOL working with new code.

thespad (approved):

PR #470, opened 2022-08-26
switch to hybrid (swag-ioncube)

aptalca:


aptalca:
locally tested

root/etc/s6-overlay/s6-rc.d/init-mod-swag-ioncube-setup/run
```diff
@@ -0,0 +1,42 @@
...
+    echo "**** Linux architecture found: x86-64 ****"
+  elif [ "${ARCH}" = "aarch64" ]; then 
+    ARCH="aarch64" 
+    echo "**** Linux architecture found: arm64 ****"
+  elif [ "${ARCH}" = "armv7l" ]; then 
+    ARCH="armv7l" 
+    echo "**** Linux architecture found: armhf ****"
+  else
+    echo "**** Unsupported Linux architecture ${ARCH} found, exiting... ****"
+    exit 1
```
thespad, line 20: Do we want to exit 1 as that will stop anything that depends on the mod from starting - i.e. all container services. Might be desired?

aptalca: I don't think that condition will ever be met as our baseimages don't support any other arches. It's one of those things that some community members like to add and I let it slide.

thespad (approved):

PR #479, opened 2022-08-26
switch to hybrid (universal-stdout-logs)

aptalca:


aptalca:
locally tested

root/etc/s6-overlay/s6-rc.d/svc-mod-universal-stdout-logs/run
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/with-contenv bash
+
+if [ -n "${LOGS_TO_STDOUT}" ]; then
+    TAIL_LOGS=$(echo "$LOGS_TO_STDOUT" | sed 's#|# #g')
+    echo "Executing: tail -F $TAIL_LOGS"
+    tail -F $TAIL_LOGS
+else
+    echo "**** Env var LOGS_TO_STDOUT is not set, sleeping ****"
+    sleep infinity
```
thespad, line 9: Instead of this you can do a `finish` file that exits 125 and that'll shut the service down and not restart it, rather than leaving it sleeping forever.

thespad (approved):

PR #480, opened 2022-08-26
switch to hybrid (universal-wait-for-internet)

aptalca:


aptalca:
locally tested

thespad (approved):

thespad (approved):
Should this be added to the deps for init-mods-end?

thespad (commented):
Should this be added to the deps for init-mods-end?

Especially as it's waiting for something - otherwise the package install step will run in parallel, which will then allow init-mods-end to trigger.

aptalca:
it's a typo. the folder `init-mods` should have been `init-mods-end`

thespad:
As in before *everything* or just before other mods? Right now it'll run immediately but will only stop mod processing (and everything after it).

aptalca:
mostly everything. The v2 version was running as `15-wait-for-internet`. I didn't see a more appropriate anchor. We should probably deprecate this

PR #502, opened 2022-09-21
Account for standard-base changes to ldap

nemchik:
https://github.com/linuxserver/docker-swag/pull/169
https://github.com/linuxserver/docker-baseimage-alpine-nginx/pull/84
https://github.com/linuxserver/reverse-proxy-confs/pull/483

Should be able to merge now (before any of the others) since it should be compatible with existing configs as well as new changes.

PR #503, opened 2022-09-22
Adjust for compatibility with standard-base

nemchik:
Find ALL files that end with `*.conf` (instead of a specific list of files). Mostly this covers the new `ldap` files (ref: https://github.com/linuxserver/docker-mods/pull/502 ), but actually is beneficial by covering anything the user could throw at it.


aptalca:
The issue with watching files is that a new file created will not be watched until a restart. Watching a folder allows for new file creation to trigger a reload

PR #506, opened 2022-09-25
Add java17-mod for code-server

christianeirich:
Please create a new branch from this PR to support java17 in code-server.


aptalca:
It's a single package install, you're probably better off using https://github.com/linuxserver/docker-mods/tree/universal-package-install

PR #508, opened 2022-09-26
add support for pip packages

aptalca:


root/etc/s6-overlay/s6-rc.d/init-mod-universal-package-install-add-package/run
```diff
@@ -1,16 +1,32 @@
...
-if [ -z ${INSTALL_PACKAGES+x} ]; then
+if [ -z "${INSTALL_PACKAGES+x}" ] && [ -z "${INSTALL_PIP_PACKAGES+x}" ]; then
     echo "**** No packages to install ****"
     exit 0
 fi
 
+if [ -n "${INSTALL_PIP_PACKAGES}" ] && [ -f /usr/bin/apt ]; then
+    echo "\
+        python3-dev \
+        python3-pip" >> /mod-repo-packages-to-install.list
```
nemchik, line 12:
```suggestion
    {
        echo "python3-dev"
        echo "python3-pip"
    } >> /mod-repo-packages-to-install.list
```

When I run 
```
echo "\
        python3-dev \
        python3-pip"
```
I get
```
        python3-dev         python3-pip
```
(a single line with a bunch of white space)

root/etc/s6-overlay/s6-rc.d/init-mod-universal-package-install-add-package/run
```diff
@@ -1,16 +1,32 @@
...
+elif [ -n "${INSTALL_PIP_PACKAGES}" ] && [ -f /sbin/apk ]; then
+    echo "\
+        python3-dev \
+        py3-pip" >> /mod-repo-packages-to-install.list
```
nemchik, line 16:
```suggestion
    {
        echo "python3-dev"
        echo "py3-pip"
    } >> /mod-repo-packages-to-install.list
```
I hate alpine for having `python3-*` and `py3-*` packages.

nemchik (changes requested):
The `echo` syntax looked a bit sus to me, and testing in a shell didn't result in the output I would have expected, so I've suggested an alternative (tested).

root/etc/s6-overlay/s6-rc.d/init-mod-universal-package-install-add-package/run
aptalca: We just escape the line endings, other than that no harm in a bunch of spaces. It's the same convention we used in every mod that installs a package. It's been thoroughly tested in production.

nemchik: Is the goal for the list file to have each package on a new line and then iterate over each line? or for each package to be separated by any kind of white space (potentially including newlines)?

I believe the current syntax can result in the file looking like this:
```
package1         package2
package3
package4 package5 package6
package7
```
If that doesn't cause issues when the file is used to install packages then it's fine (just bugs my OCD a bit).

aptalca: The only goal is to have a file that lists a bunch of package names. Empty spaces and newlines are fine. 

We don't care about how it looks because no one is ever supposed to even see that file. It's entirely ephemeral and internal as it gets created and deleted during init. It's just a temporary storage mechanism so different init scripts and mods can compile a list of packages and then they can be installed all at once.

It fully complies with KISS 😄

nemchik: I just went and double checked https://github.com/linuxserver/docker-baseimage-alpine/blob/0fa7d8bbf2a7791afbd51aa035b48d228aa639e2/root/etc/s6-overlay/s6-rc.d/init-mods-package-install/run#L5 (sorry it's just been a minute since I looked at it) and realize we are specifically replacing newlines with spaces and then making an array of values by splitting the spaces in the string. So basically anything we throw at it should work 😁

nemchik (approved):

PR #526, opened 2022-11-19
code-server-dotnet: update versions filter

Fronti:
Due to changes in releases-index.json the script won't be able to find any releases anymore. 

I updated the filters to reflect these changes, which are also described [here](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#servicing). 
tl;dr: "active" is, well, actively supported. "active" releases go into "maintenance" 6 months before EOL and from that point on only receive security updates until they reach EOL.


Fronti:
My bad, updated it

aptalca (approved):
Excellent, thanks

PR #529, opened 2022-11-29
Fix some typos in README.md

trolley:


trolley:
Should I be requesting a review from anyone in particular?

PR #557, opened 2023-01-15
Check if cron running before starting

mtongnz:
The nginx container would have multiple instances of cron running causing jobs to fire multiple times.

This will exit the run scripts if cron or crond is already running


aptalca:
An `exit` will put this service in a start loop.

Perhaps `sleep infinity` would be better

nemchik:
FYI this is being looked at from another angle (a bit more globally) https://github.com/linuxserver/docker-mods/pull/736 if you want to subscribe, but we'll also likely be posting to https://info.linuxserver.io

PR #564, opened 2023-01-21
Create with-contenv alias

nemchik:


thespad:
Have you run this locally to make sure it does what it's supposed to? We don't have a good way to automate testing because of how it gets pulled into downstream bases and it's a hassle to revert or change if something breaks.

nemchik:
I just added 8b3658790327fbd2bb1668501b1356e9388c3421 and confirmed that mounting
```yaml
      - "${DOCKERCONFDIR}/test-container/docker-mods:/docker-mods"
```
does in fact work as expected. I ran it as it is here in the PR, and also ran it slightly modified to force it to replace `/usr/bin/with-contenv` with the one from the mod init. Both behaved as expected.

If the image has `/usr/bin/with-contenv` the logic is skipped and a message is logged. If `/usr/bin/with-contenv` does not exist it is created and works as expected.

Alternatively, we could consider clobbering `/usr/bin/with-contenv` when it is included in the image. The logic would still only do this if `/command/with-contenv` exists, so it shouldn't break if it were run on an older s6v2 image (even though we don't have this mod init running on any older images).

thespad:
So right now there are three possible paths:

Image gets the contenv symlink from /command to /usr/bin
Image copies our version on build into /usr/bin, replacing the symlink
Mods script creates our version on init, replacing the symlink or copied version.

So we could clobber, which would ensure that if the file is removed from the repo we still create ours - but as long as we're doing `-f` to check and not something that will match on symlinks, it should overwrite anyway.

nemchik:
I was incorrect in describing the logic I wrote, and the logic was wrong. I have pushed dad34393d33daa565e2606d9aafe1162401eb918 to fix it. The logic is now:

If `/command/with-contenv` (from s6v3) does NOT exist, skip creating our alias for `/usr/bin/with-contenv` (which relies on `/command/with-contenv`).

This COULD mean the user ends up with no `/usr/bin/with-contenv` at all if it wasn't put in place by s6 (either s6v2 binary or s6v3 symlink) or we didn't bake it into the base image (with s6v2 we renamed the original binary to `/usr/bin/with-contenvb` and shipped a custom alias, with s6v3 so far we are just replacing the symlink with our alias). This should never happen though. We are only using this mod init script in s6v3 images that should include `/command/with-contenv` and so should trigger the logic to create the alias, and clobber anything in the way of creating the alias.

So if this merges, `/usr/bin/with-contenv` will always be our alias from this script when it is used on an image with s6v3.

thespad (approved):
LGTM

Tested locally with and without the existing with-contenv file being copied, all seems to be working as expected.

Issue #570, opened 2023-01-28
swag-dashboard : Flooding DNS server with requests with default swag config


hadjilucasL:
Hi,

I got a warning on my DNS server about many concurrent requests from the host on which SWAG is hosted.

Like many attempts to resolve services like:

```
adguard
adguard.mydomain.com
adminer
adminer.mydomain.com
adminmongo
adminmongo.mydomain.com
...
```

At first I thought something was compromised and someone was scanning my local network but upon investigation it turns out this behaviour is coming from the SWAG Dashboard.

The culprit is down to the code [here](https://github.com/linuxserver/docker-mods/blob/6e6ddfb166974561599e55ca3ea94591990e1fd1/root/dashboard/swag-proxies.py#L13). 

From my experience when first setting up SWAG, the default config included all these [reverse proxy config samples](https://github.com/linuxserver/reverse-proxy-confs) that are only activated if you rename the suffix from `.conf.sample` to `.conf`. These are placed in the ` /config/nginx/proxy-confs` directory. This is indicated in the SWAG documentation as well [here](https://docs.linuxserver.io/general/swag#preset-proxy-confs).

From looking at the code, SWAG Dashboard globs all files in the nginx conf directory, including the samples and as a result it spams a lot of concurrent connections that might otherwise look like someone malicious is probing.

This behaviour also gave rise to issue #501.

The suggestion is to rework the code as to at least only glob files ending in `.subdomain.conf` or `.subfolder.conf`.

aptalca:
It's not a bug. It is intended behavior

hadjilucasL:
Thanks for the response @aptalca . 

I am not quite sure I understand how is this intended behaviour ? The files ending in `.sample` are not being considered as active in nginx. Yet they are present as per default configuration to facilitate users. So anyone enabling SWAG Dashboard will inevitably be making a number of connection attempts to these yet-to-be-enabled configurations. Why add these in the applications being scanned ? 

Furthermore some DNS servers do impose warnings on the number of concurrent connections (e.g. pihole is set at 150). The current code will launch a large number of threads without rate limitation and attempt to connect to all sample reverse proxy configurations. There are currently 270 samples so it can add up every time you visit the SWAG Dashboard page.

aptalca:
A feature of the dashboard is to show you if you have any of those known services running but not reverse proxied (perhaps not properly activated).

The dns warning is a false positive so not something that needs fixing on the dashboard's side.

[...]

Issue #501, opened 2022-09-18
[swag-dashboard] Unknown applications showing up in Swag Dashboard


Unlimited0012:
Two unknown applications showed up in Swag Dashboard.

There are no active proxy confs for either of these applications, and there are no set subdomains for them in my yml file.

The applications are:
- http://audiobookshelf:80/
   - Available: YES
   - Proxied: NO
   - Location: (this field is blank)
- http://tasmobackup:80/
   - Available: YES
   - Proxied: NO
   - Location: (this field is blank)

I've seen one other report on Reddit with the user reporting "tasmobackup" and "documentserver" randomly showing up similarly.

aptalca:
That just means your dns resolves both of those hostnames from inside SWAG and something responds to requests on those IPs and port 80. But they are not reverse proxied by SWAG.

Unlimited0012:
@aptalca thank you.

Could you help me understand? 
Does this mean that there are services on my server broadcasting those hostnames?

aptalca:
The only things for sure are (iirc) 
1) your dns (or rather docker dns) is resolving those hostnames to some IP(s) and
2) those IPs are listening and responding on port 80

MilesTEG1:
I was having the same issue with the exact same domain. But I was not using those.
I see that when I delete the .sample file corresponding to those domains, they were gone.

But since the yesterady update of swag, the dashboard doesn't show any proxies now:
![image](https://user-images.githubusercontent.com/29298312/191983202-4270acc9-e5d5-404d-80f0-c9efbe2de5ce.png)

Unlimited0012:
@MilesTEG1 Evidently the main issue of unknown applications showing up was sorted out on the linuxserver discord. I forget the exact specifics, but essentially it's not threatening.

I haven't noticed my proxies missing as yours are though.

Shiwigy:
I also have the exact same problem, it's an easy fix if you edit /dashboard/swag-proxies.py line 13

FROM
    file_paths = glob.glob("/config/nginx/**/**", recursive=True)
TO
    file_paths = glob.glob("/config/nginx/**/*.conf", recursive=True)

It seems it picks up every available file by default, with the change above it only process the files "enabled" *.conf and not the *.conf.sample files

MilesTEG1:
Hello,
I forgot I had post here.
[...]

PR #571, opened 2023-01-28
swag-dashboard : Provide status on only enabled proxy configs

hadjilucasL:
This PR is an attempt to resolve the issues identified in #501 and #570.

Basically the active reverse proxy configurations being considered are now explicitly only in the files with the following extensions:

```
*.subdomain.conf
*.subfolder.conf
```

and only in the default SWAG reverse proxy configuration directory which is `/config/nginx/proxy-confs`. This is based off the official documentation [here](https://docs.linuxserver.io/general/swag#preset-proxy-confs) and [here](https://github.com/linuxserver/reverse-proxy-confs#configure-your-default-site-config) that guides users on how to enable different proxy configurations.

I was not sure why conf files under `/config/nginx/site-confs/` and `/etc/nginx/http.d/` where being analysed as you would not normally expect these to contain reverse proxy configurations. I could also not determine where these apps marked as `auto-proxy` where being surfaced to the user. If someone can provide an indication as to how these are used I can restore the relevant lines of code.

### DIFF
--- a/root/dashboard/swag-proxies.py
+++ b/root/dashboard/swag-proxies.py
@@ -8,12 +8,11 @@ from pathlib import Path
 import urllib3
 
 
-def get_proxy_conf_file_paths():
+def get_reverse_proxy_conf_file_paths():
     """Get all the active nginx proxy configuration files"""
 
     file_paths = []
 
-    # include reverse proxy configs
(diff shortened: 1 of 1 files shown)


aptalca:
Current behavior is the intended behavior

PR #576, opened 2023-02-05
Add accept headers + Test tarballs

thespad:


aptalca (approved):

thespad:
Adding the Accept headers changes the response from Dockerhub (to the same format as GHCR), which will break downloading the mods. Changes from this PR are rolled into https://github.com/linuxserver/docker-mods/pull/577 along with a fix for the changed manifests.

PR #584, opened 2023-02-16
Update Readme for correct UK country code: GB

Tsunami2056:
It seems the actual geoip2_data_country_iso_code country code for the UK is GB (great britain).


thespad:
You are correct, although almost all systems that use the ISO 3166 country codes treat UK and GB as the same (and UK is specifically reserved, so it cannot be assigned to any other country).

PR #609, opened 2023-03-25
Check existing permissions before calling chown in lsiown

kaysond:
Blind chown-ing means the error message is shown even if the permissions are already correct. For example, if you have the directory mounted over nfs with root squashed, chown will always fail. But the permissions might be set correctly so it doesn't actually matter, and there's no need to display the message.

The simplest thing would just be to check the target for permissions, and if they're already set, to skip the chown call. We'd have to check for the `-R` argument and do it recursively, if necessary. There could be other arguments, but briefly scanning through [this search](https://github.com/search?q=org%3Alinuxserver+lsiown&type=code&p=1) showed only the one.

Another option might be to run `find` and only actually call chown on files that don't have the correct permission. This might be better for performance on large filesets, and if it failed, it would be a true failure. It would also require some argument parsing.

### DIFF
--- a/docker-mods.v3
+++ b/docker-mods.v3
@@ -103,6 +103,7 @@ create_lsiown_alias() {
     # intentional tabs in the heredoc
     cat <<-EOF >/usr/bin/lsiown
 	#!/bin/bash
+	# Check ownership here
 	chown "\$@" || printf '**** Permissions could not be set. This is probably because your volume mounts are remote or read-only. ****\n**** The app may not work properly and we will not provide support for it. ****\n'
 	EOF
     chmod +x /usr/bin/lsiown


kaysond:
I wanted to get some feedback on the implementation (and issue in general) before going forward. Looks like @thespad might be the right person for that?

aptalca:
> Blind chown-ing means the error message is shown even if the permissions are already correct

That is indeed intended behavior. We don't support read only or remote mounts for config. All of our init processes assume the container has full rw access to the config folder. If that's not the case, we don't provide support and making that explicit through log messages saves us a lot of time during support and troubleshooting.

kaysond:
The read only I understand, but why not remote mounts, if the init processes just care about write access?

aptalca:
Database corruption is one reason

kaysond:
> Database corruption is one reason

Ah. Yeah. It annoys me that sqlite is stubborn about that even though every mainstream nfsv4 implementation does locking properly now.

Would you guys take a PR for a regular docker mod to make changes to `lsiown`? So it wouldn't be default behavior, but for people like me with more complicated storage setups, we could, for example, get a readout of files with wrong permissions instead of just the one error message.

thespad:
A regular mod would run after all the other init steps, which is after the built-in chowning will have already been completed.

You could mount your own `docker-mods` with its own `lsiown` alias, but then you'd have to manually track any upstream updates to the file so that remain in sync.

The problem is that the alias as implemented is literally just chown with an echo on error, which means you can't just add permissions detection without parsing out any arguments that could have been passed to it to identify the files/folders it's been called against. And you can't fundamentally change the behaviour of the alias without having to change every downstream image which uses it (basically all of them at this point).

kaysond:
> A regular mod would run after all the other init steps, which is after the built-in chowning will have already been completed.

That's just by convention, right? From what I can see, the `docker-mods` script just untars a container layer blob and copies the contents over `/`. And it does that **after** it installs `lsiown`. So if my mod is just a `/usr/bin/lsiown` file, it would overwrite the default one. 

> The problem is that the alias as implemented is literally just chown with an echo on error, which means you can't just add permissions detection without parsing out any arguments that could have been passed to it to identify the files/folders it's been called against. And you can't fundamentally change the behaviour of the alias without having to change every downstream image which uses it (basically all of them at this point).

I don't think it would be too hard to parse the arguments in a custom version since you're already using bash and the replacement could just use the `getopts` builtin, especially since only `-R` is used so far anyways.

In essence, what I'm trying to accomplish is something like `find $PARSED_PATH ! -user $PARSED_USER -o ! -group $PARSED_GROUP -print -exec chown $CHOWN_ARGS {} + || echo "Error Message"`

It also occurred to me that there's no checking of the actual permissions. It's admittedly pathological, but the file could have no read or write perms for anyone. More likely you might find a scenario where the user/group has no write permissions. Given that there's only one chown option being used so far, I think there's an opportunity to expand the rw/owner checks, if that's something you guys want to do.

kaysond:
Not stale

PR #613, opened 2023-04-01
Fix version links and change the mod to use the alpine package

quietsy:
- Changed to the new path of the version script under `/etc/s6-overlay/s6-rc.d/init-version-checks/run`
- Updated the version links to point to the new locations under the alpine nginx base
- Changed to use the goaccess alpine package instead of building it
- Removed the old unused rc.d script


root/etc/cont-init.d/98-dashboard-config
```diff
@@ -1,30 +0,0 @@
-#!/usr/bin/with-contenv bash
```
quietsy, line 1: @thespad This isn't needed anymore, right?

aptalca (approved):

PR #630, opened 2023-04-20
Add /dev/kfd device

CrackedP0t:
According to https://github.com/RadeonOpenCompute/ROCm-docker, ROCm also requires /dev/kfd. On my machine with a Vega 56, OpenCL wouldn't work without it.


aptalca:
@PascalMinder can you confirm?

PascalMinder:
I don't have a Vega 56 graphics card to verify. I'll look into it on the weekend.

nyanmisaka:
As per the docs from AMD, `/dev/kfd` is required by the ROCm runtime on Linux.

https://github.com/RadeonOpenCompute/ROCm-docker#docker-hub

PR #661, opened 2023-05-18
code-server-ssl update/add workflows

aptalca:


Roxedus:
Is this supposed to drop openvscode support?

aptalca:
Yeah, I realized those arguments aren't supported by openvscode-server

Roxedus (approved):

PR #693, opened 2023-05-18
universal-docker update/add workflows

aptalca:


.github/workflows/BuildImage.yml
```diff
@@ -1,78 +1,47 @@
...
-          docker push ghcr.io/${ENDPOINT}:${BASEIMAGE}-${MODNAME}-${{ github.sha }}
-          docker push ghcr.io/${ENDPOINT}:${BASEIMAGE}-${MODNAME}
+          echo "GITHUB_REPO=${{ env.GITHUB_REPO }}" >> $GITHUB_OUTPUT
+          echo "ENDPOINT=${{ env.ENDPOINT }}" >> $GITHUB_OUTPUT
+          echo "BASEIMAGE=${{ env.BASEIMAGE }}" >> $GITHUB_OUTPUT
+          echo "MODNAME=${{ env.MODNAME }}" >> $GITHUB_OUTPUT
+          # **** If the mod needs to be versioned, set the versioning logic below. Otherwise leave as is. ****
+          COMPOSE_RELEASE=$(curl -sX GET "https://api.github.com/repos/docker/compose/releases/latest" | awk '/tag_name/{print $4;exit}' FS='[""]' | sed 's|^v||')
+          DOCKER_RELEASE=$(curl -sX GET "https://api.github.com/repos/moby/moby/releases/latest" | awk '/tag_name/{print $4;exit}' FS='[""]' | sed 's|^v||')
+          MOD_VERSION="${DOCKER_RELEASE}-${COMPOSE_RELEASE}"
```
thespad, line 25: What happens if they end up with a `-` in a release name somewhere along the line? Maybe use a `|` or something else that's never going to realistically be there?

aptalca: That character is not allowed in docker tags. Only hyphen, underscore and period are. I know not ideal, but we used the dash separator for this mod for years and it hasn't been an issue. I'm willing to continue taking the risk. If it fails at some point, we can fix it then.

thespad: Oh I see what you mean, I had forgotten it would be tagged with it.

thespad (approved):

Issue #727, opened 2023-07-09
[Jellyfin rffmpeg mod] Native WoL from outside NAT.


a-priestley:
For those with more complicated network topologies, broadcast magic packet may not be enough.
My set up at the moment requires sending the packet to the WAN side of the switch, targeting a port mapped to the specific machine I want to wake, along with that machines MAC address.
Is it feasible to implement?

aptalca:
@junkman690

a-priestley:
I made a small edit to the `root/usr/bin/wol_rffmpeg/wol` script, with some extra environment variables for wol port and IP. It seems to work now but may require more testing.
Do you accept PRs?

PR #728, opened 2023-07-10
Jellyfin rffmpeg NAT WoL

a-priestley:
https://github.com/linuxserver/docker-mods/issues/727


aptalca (changes requested):
@a-priestley thanks for the PR
Can you add bits to the readme to clarify that these two new vars are optional and that they should be either not set or left blank if not needed?

a-priestley:
@aptalca Done

aptalca (approved):

PR #737, opened 2023-07-31
Make rffmpeg universally compatible

JeWe37:
rffmpeg is mostly used with jellyfin, however particularly with the introduction of joshuaboniface/rffmpeg#70 there are other uses, including with other linuxserver containers. Therefore it makes sense to make this image universal. Support has been added for alpine base images.

The branch and image may need to be renamed after this is merged.

WIP for now, I'd like to verify everything is working with the pipeline images.

### DIFF
--- a/root/etc/s6-overlay/s6-rc.d/init-mod-jellyfin-rffmpeg-add-package/run
+++ b/root/etc/s6-overlay/s6-rc.d/init-mod-jellyfin-rffmpeg-add-package/run
@@ -1,10 +1,24 @@
 #!/usr/bin/with-contenv bash
 
-for package in iputils-ping openssh-client python3-click python3-yaml wakeonlan; do
-    if ! dpkg -s "${package}" >/dev/null 2>&1; then
-        PACKAGES="${package} ${PACKAGES}"
+## Ubuntu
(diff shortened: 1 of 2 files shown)


JeWe37:
Just a minor fix to use the right kind of wake on lan, rest should be working.

aptalca:
Thanks for the PR, but it wouldn't work unless the branch was moved as well

Also, I'm not in favor of this simply because this is not a mod that our team maintains. And so far none of the team members are interested in taking over maintenance (only 1 uses jellyfin I believe) so we would not want to promote this as a universal mod.

aptalca:
With that said, what I'd be ok with is, adding just the bits necessary for installing packages on alpine, and leave the mod as is. That way, we're not promoting it as a universal, but users can still use it for other containers at their own risk.

JeWe37:
Makes sense, I don't really care what the image is advertised as so that's fine by me. Updated everything accordingly, not it's essentially only installing the packages required and using the right wol command name.

JeWe37:
Fixed the conflicts with the other merged PR still, didn't notice those before.

aptalca (approved):

PR #754, opened 2023-08-29
Abort if bouncer version can't be fetched.

thespad:


Roxedus (commented):
Would it be better to fall back to todays latest version?

thespad:
I don't want to fall back because it means people could end up running old versions without knowing it. Also if the first github fetch fails it's probably going to fail to download the bouncer files too.

PR #756, opened 2023-08-30
swag: swag-dashboard add rotated logs preloading

FoxxMD:
Ingests rotated/archived nginx log files if they exist and goaccess has not yet persisted any data.

The existing configuration for goaccess stores persisted data to `/tmp` and lsio only provides the most current `access.log` by default. 

This means goaccess only has access to a small window (whatever is contained in `access.log`) in a fresh container, even if there are existing rotated/archived logs in the nginx logs directory. This window gets larger the longer that container runs as goaccess persists data BUT this is wiped if the container is re-created.

This addition to the dashboard mod, after checking existing persisted data does not exist (likely on a fresh container), attempts to "preload" rotated/archived data into goaccess so that a user can fully leverage all their nginx logs.


quietsy:
Thanks for the PR!

This behavior is by design, it takes a very long time for some users to even load just the single `access.log`, and others have modified logrotate to keep a lot of logs.
The current behavior was chosen for being the least problematic and because the most common use-case is to check for recent activities, not as a historical archive.

Besides that, you also need to change `keep-last 30` in `goaccess.conf` if you want goaccess to keep data for longer.
Once the container is created it will only process `access.log` if you open the dashboard, if you don't open the dashboard for 2 weeks you'll have "holes" in the data unless you recreate the container since logs may have rotated and nothing processed them in the background.
A full-blown implementation will need to have a background service ingesting the logs regularly and persisting the goaccess storage on the host, it should probably be a standalone container.

If you want to keep it inside the mod, I suggest going with your current customization script but running it in cron every hour, mounting `/tmp` on the host to persist it, and mounting `goaccess.conf` on the host to modify `keep-last`.

PR #760, opened 2023-09-03
Allow user to specify which version of rffmpeg to download

Dominaezzz:
I've been using this mod for the past few months and encountered breaking changes twice, without warning.
Unfortunately every time the jellyfin container is restarted, the latest rffmpeg is downloaded and I'm at the mercy of whatever changes have been made since I last ran.
I prefer my upgrades to always be intentional (not automatic) and for any changes to be acknowledged at a time I'm prepared to deal with them.

This PR add the ability to set the specific commit to download rffmpeg from.
This way I can opt in to changes when I chose to.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ ARG MOD_VERSION=master
 RUN \
   echo "**** grab rffmpeg ****" && \
   mkdir -p /root-layer/usr/local/bin/ && \
-  curl -o \
+  curl -fo \
     /root-layer/usr/local/bin/rffmpeg -L \
     "https://raw.githubusercontent.com/joshuaboniface/rffmpeg/${MOD_VERSION}/rffmpeg" && \
   chmod +x /root-layer/usr/local/bin/rffmpeg
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,24 @@
 # syntax=docker/dockerfile:1
 
+FROM ghcr.io/linuxserver/baseimage-alpine:3.17 as buildstage
+
+ARG MOD_VERSION=master
+
+RUN \
+  echo "**** grab rffmpeg ****" && \
+  mkdir -p /root-layer/usr/local/bin/ && \
+  curl -o \
+    /root-layer/usr/local/bin/rffmpeg -L \
+    "https://raw.githubusercontent.com/joshuaboniface/rffmpeg/${MOD_VERSION}/rffmpeg" && \
+  chmod +x /root-layer/usr/local/bin/rffmpeg
+
+# copy local files
+COPY root/ /root-layer/
+
+## Single layer deployed image ##
 FROM scratch
 
 LABEL maintainer="junkman690"
 
-# copy local files
-COPY root/ /
+# Add files from buildstage
+COPY --from=buildstage /root-layer/ /
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,11 +2,12 @@
 
 FROM ghcr.io/linuxserver/baseimage-alpine:3.17 as buildstage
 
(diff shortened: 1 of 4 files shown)


aptalca:
Instead of telling it to download a specific commit during runtime, it would be better to have it build the latest commit as versioned and enable an external trigger like we have for a few of the mods: https://github.com/linuxserver/docker-mods/blob/transmission-floodui/.github/workflows/BuildImage.yml#L23

That way all the versioning would be handled build time, no dependency on an external service during runtime, and the user can pin the docker image tag to whichever version they like.

Dominaezzz:
> and the user can pin the docker image tag to whichever version they like.

What does this look like? I haven't seen the mod syntax with any versioning info in it.

thespad:
The same as any docker image really, e.g.

DOCKER_MODS=linuxserver/mods:somemod-v1.2.3

aptalca:
https://hub.docker.com/r/linuxserver/mods/tags?page=1&name=powershell

Dominaezzz:
Just to be clear, all I have to do is copy that line and make it fetch the latest commit hash from GitHub Yeah?

Dominaezzz:
Is this acceptable? (See [latest commits](https://github.com/joshuaboniface/rffmpeg/commits/master))
```
$ curl -s https://api.github.com/repos/joshuaboniface/rffmpeg/commits/master | jq -rc '.sha'
a2422187b10f9bd9125d837cc60d3aa066cbcbe1
```

This being the only change required.
`MOD_VERSION=$(curl -s https://api.github.com/repos/joshuaboniface/rffmpeg/commits/master | jq -rc '.sha')`

aptalca:
The line I linked will handle the docker image tagging. You also need to modify the Dockerfile to download the asset (with the same versioning logic) during buildtime. Just like this one: https://github.com/linuxserver/docker-mods/blob/transmission-floodui/Dockerfile

Then you can remove the curl download step from the s6 init file.

aptalca:
> Is this acceptable?

Yes, that looks fine for the versioning logic

Dominaezzz:
I think I understand what needs doing. I just have some questions.

1. Is the multi stage build a requirement?
2. Is `ARG MOD_VERSION=master` acceptable? (Calculating the default programmatically seems excessive)
3. I don't understand the significance of the `/root-layer/themes/flood-for-transmission` folder in the build stage. rffmpeg needs to be in `/usr/local/bin/rffmpeg` at the end of the day. Does this mean I need to curl rffmpeg to `/root-layer/usr/local/bin/rffmpeg` in the build step?

aptalca:
1. Yes because the final mod has to be a single layer docker image
2. That should be fine
3. No, for curl, you'd need a `RUN` statement which is a layer. Then you'd have a `COPY` statement for the init files, which would be another layer. The multi stage allows you to prepare all necessary files in one path in a build stage, so you can do just a single `COPY --from=buildstage` directive in the final scratch image, resulting in a single layer.

Dockerfile
```diff
@@ -1,8 +1,24 @@
 # syntax=docker/dockerfile:1
 
+FROM ghcr.io/linuxserver/baseimage-alpine:3.17 as buildstage
+
+ARG MOD_VERSION=master
+
+RUN \
+  echo "**** grab rffmpeg ****" && \
+  mkdir -p /root-layer/usr/local/bin/ && \
+  curl -o \
```
aptalca, line 10: Please make this `curl -fo` so the build fails when GitHub is not reachable or errors out

Dominaezzz:
Unsure why `MOD_VERSION` isn't being set. It's like CI is ignoring the change I made.

aptalca (approved):

PR #766, opened 2023-09-21
Add the parameter SHELL_NOLOGIN

barhun:
This PR adds the functionality to disable user login shell. Some people, like myself, may need only tunneling and want to disable shell access which they'll never use.


root/etc/s6-overlay/s6-rc.d/init-mod-openssh-server-ssh-tunnel-setup/run
```diff
@@ -4,3 +4,10 @@
 sed -i '/^AllowTcpForwarding/c\AllowTcpForwarding yes' /etc/ssh/sshd_config
 sed -i '/^GatewayPorts/c\GatewayPorts clientspecified' /etc/ssh/sshd_config
 echo "TcpForwarding is enabled"
+
+[[ "$SHELL_NOLOGIN" == "true" ]] && {
```
aptalca, line 8: Please change this to an if block, because this format results in a non zero exit if the first statement is not true and can block container start by tripping up s6 overlay

aptalca (approved):

PR #771, opened 2023-10-12
detect if firefox is installed in the container and download the langs

thelamer:
This adds logic for firefox lngpacks as they have their own dedicated extensions not tied to sys level packages.


thelamer:
linking https://github.com/linuxserver/docker-firefox/issues/34

thespad (approved):

PR #778, opened 2023-10-30
Add support for new `swag_server_custom_directive` label

### LINKED ISSUE #764: [SWAG Auto-proxy] Add support for geoblocking
The current dbip or maxmind mods for swag allow geoblocking support by adding explicit lines in application's individual config file.

When using auto proxy, the user does not have individual config files.

Suggestion:
Add a new label `swag_geoblock` with possible values: `none; allow; deny` to automatically include the relevant snippets for blocking based on the dbip or maxmind mods.

aptalca:
- Inject a custom directive into the server block via `swag_server_custom_directive` label
- Make compatible with upcoming cron changes
- Sanitize `swag_url` sed replace

closes #764

### DIFF
--- a/root/etc/s6-overlay/s6-rc.d/init-mod-swag-auto-proxy-setup/run
+++ b/root/etc/s6-overlay/s6-rc.d/init-mod-swag-auto-proxy-setup/run
@@ -12,8 +12,7 @@ rm -rf /auto-proxy
 mkdir /auto-proxy
 
 if ! grep -q "/app/auto-proxy.sh" /config/crontabs/root; then
-    echo "*       *       *       *       *       /app/auto-proxy.sh" >> /config/crontabs/root
-    cp /config/crontabs/root /etc/crontabs/root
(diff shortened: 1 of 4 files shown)


aptalca:
tested locally

root/etc/s6-overlay/s6-rc.d/init-mod-swag-auto-proxy-setup/run
```diff
@@ -12,8 +12,7 @@ rm -rf /auto-proxy
 mkdir /auto-proxy
 
 if ! grep -q "/app/auto-proxy.sh" /config/crontabs/root; then
```
nemchik, line 14:
```suggestion
if ! grep -q "/app/auto-proxy.sh" /etc/crontabs/root; then
```

nemchik (changes requested):
Aside from the one change suggestion, swag runs all other cron as abc now instead of root. there's no issue if this needs to run as root, but if it can be run as abc there would be some changes needed.

nemchik (approved):

PR #808, opened 2023-12-29
code-server: rust initial release

totraku:


totraku:
I am not sure why MOD_VERSION is not being set on the PR build. It surely is present on my repo builds.

aptalca:
Because you're not an authorized user for this repo, GitHub won't run the workflow from your branch. It's a security feature.

aptalca (changes requested):
Please set the version in the Dockerfile as well so local builds (and PR builds) succeed (ie. https://github.com/linuxserver/docker-mods/blob/code-server-julia/Dockerfile#L8-L12 )


Also, how come there is no aarch64 support? Pretty sure upstream supports it.

totraku:
> Also, how come there is no aarch64 support? Pretty sure upstream supports it.
We can add it, the image will get a little bigger, although they are all removed after installation.
The other way would be to download rustup in the image and then do the installation using the s6 script, so rustup decides what arch to download. While this approach would provide support for a lot of architecturea, I thought this is not the preferred one. Let me know what you think

totraku:
Ok so I added aarch64, and also included the version fetching for MOD_VERSION in the Dockerfile

totraku:
Right now the image size is 1.576GB and it contains x86_64-unknown-linux-musl x86_64-unknown-linux-gnu aarch64-unknown-linux-musl aarch64-unknown-linux-gnu. As previously mentioned, one solution could be to add different tags for the C standard library `gnu` / `musl` and architecture `x86_64` / `aarch64`.
Those images would have a quarter of the size, so roughly 400 MB.

aptalca:
Single tag for a mod is preferred. Our current mod infra doesn't support multi tags.

Question, though: Both code-server and openvscode-server are ubuntu based. Does this mod need the musl packages?

If your intention for this is to be used by other images as well, then you can keep the musl bits but name the mod `universal-rust` instead.

If it's for code-server/openvscode-server, then there is no need for the musl bits

totraku:
We can just leave the musl images out. Personally, I have no other use-case than code-server. As you mentioned, those images being ubuntu based, make the musl part obsolete. At that point the image will be roughly, 700mb. Is it acceptable, or should we introduce some ARCH tag, that would half the size once more?

aptalca:
That's perfectly acceptable. We have larger mods than that (dind and dotnet). I'd rather rely on downloading the packages from upstream during build time because ghcr (Azure) and docker hub (aws) are much better to download from compared to whatever the upstream projects may use. Plus, the images are pulled from ghcr (or docker hub) to begin with so a working connection to them is already a prerequisite for the users whereas other upstream sources may be blocked (pihole and such).

totraku:
Ok, I removed all the musl binaries.

Dockerfile
```diff
@@ -1,8 +1,26 @@
 # syntax=docker/dockerfile:1
+FROM ghcr.io/linuxserver/baseimage-alpine:3.19 as buildstage
+
+ARG MOD_VERSION
+
+RUN apk add --no-cache curl jq && \
```
aptalca, line 6: Our 3.19 baseimage includes jq and curl so this line can be removed

root/etc/s6-overlay/s6-rc.d/init-mod-code-server-rust-install/run
```diff
@@ -0,0 +1,39 @@
...
+  echo "+++ unpacking rust tar +++"
+  tar -xzf "$RUST_BINARY_ARCHIVE_PATH" -C "$RUST_BINS_PATH" --strip-components=1
+
+  echo "+++ installing rust +++"
+  "$RUST_BINS_PATH/install.sh"
+  rm -rf "$RUST_BINS_PATH"
+
+  echo "+++ adding build tools +++"
+  if grep_os_release "debian"; then
+    echo "build-essential" >>"$PACKAGES_TO_INSTALL_LIST"
```
aptalca, line 32: Please add a check to see if it's already installed like here: https://github.com/linuxserver/docker-mods/blob/code-server-python3/root/etc/s6-overlay/s6-rc.d/init-mod-code-server-python3-add-package/run#L3

Mod tarballs can be updated even when the container is restarted, not just recreated so there are cases where the mod tarball exists, but the repo packages are already installed.

totraku: https://github.com/linuxserver/docker-mods/pull/808/commits/17a69de8dac0032eacf7a1c5d84d93f510f92019#diff-22bc35d6f7fcd5baecdae1ae65c1d1b48f027b17e5583a0a46c490a5fe155a84R5-R9

Dockerfile
totraku: done

root/etc/s6-overlay/s6-rc.d/init-mod-code-server-rust-install/run
aptalca: That won't catch the scenario I mentioned above. Actually, that part should be removed because it will prevent a rust update when the init downloads a newer version mod on a restart.

totraku: Ok, I think you know what you mean, but dpkg check won't work, I think the current behavior should catch this, case. I changed the message:
https://github.com/linuxserver/docker-mods/blob/a180cf76cf81f69950590012103534532fb25552/root/etc/s6-overlay/s6-rc.d/init-mod-code-server-rust-install/run#L27

totraku: The golang mod, does it this way too:
https://github.com/linuxserver/docker-mods/blob/code-server-golang/root/etc/s6-overlay/s6-rc.d/init-mod-code-server-golang-add-package/run

aptalca: To further clarify, there are two things that need installed, rust package, and the repo package. They need separate checks because the former can be updated and may need a reinstall on a container restart, whereas the latter only needs to be installed after a recreation.

So it's prudent to have a check for the bins for the rust package, and if there are, install them (could mean it was never installed, or was installed but has an update, either way, we need to install the bins). And have a separate dpkg check for the repo package and add to package install list as needed.

Golang mod also needs to be updated with the repo package check.

totraku: Sorry, it has been a long day. I totally forgot about the packages. So it should suffice to make the package check outside independently of the rust package check, right?

totraku: e.g.:

`#!/usr/bin/with-contenv bash

set -e

RUST_BINS_PATH="/rust-bins"
PACKAGES_TO_INSTALL_LIST="/mod-repo-packages-to-install.list"

ARCH=$(uname -m)
SUPPORTED_ARCHS=("x86_64" "aarch64")

if [[ ! " ${SUPPORTED_ARCHS[@]} " =~ " ${ARCH} " ]]; then
  echo "**** unsupported architecture: $ARCH ****"
  exit 1
fi

RUST_BINARY_ARCHIVE_PATH="$RUST_BINS_PATH/rust-$ARCH-gnu.tar.gz"

if ! dpkg -l | grep build-essential >/dev/null; then
  echo "**** installing build-essential ****"
  echo "build-essential" >>"$PACKAGES_TO_INSTALL_LIST"
else
  echo "**** build-essential already installed ****"
fi

if [[ -f "$RUST_BINARY_ARCHIVE_PATH" ]]; then
  echo "**** unpacking rust tar ****"
  tar -xzf "$RUST_BINARY_ARCHIVE_PATH" -C "$RUST_BINS_PATH" --strip-components=1

  echo "**** installing rust ****"
  "$RUST_BINS_PATH/install.sh"
  rm -rf "$RUST_BINS_PATH"
else
  echo "**** rust bin does not exist ****"
fi
`

root/etc/s6-overlay/s6-rc.d/init-mod-code-server-rust-install/run
```diff
@@ -26,13 +29,6 @@ if [[ -f "$RUST_BINARY_ARCHIVE_PATH" ]]; then
...
-
-  echo "**** adding build tools ****"
-  if grep_os_release "debian"; then
-    echo "build-essential" >>"$PACKAGES_TO_INSTALL_LIST"
-  else
-    echo "gcc" >>"$PACKAGES_TO_INSTALL_LIST"
-  fi
 else
-  echo "**** rust already installed ****"
+  echo "**** rust bin does not exist ****"
```
aptalca, line 33: Last nitpicky comment I promise :-)
While this log message is technically correct from a coding perspective, it would be confusing to an end user. Can you change it to something like `latest rust is already installed`? Currently it sounds like something unexpected happened whereas it really means no install or update is necessary.

root/etc/s6-overlay/s6-rc.d/init-mod-code-server-rust-install/run
aptalca: Yup, this looks great

root/etc/s6-overlay/s6-rc.d/init-mod-code-server-rust-install/run
totraku: No problem 
I am thankful for your detailed explanation and resilience :smile:

aptalca (approved):

totraku:
One more question, how are the changes merged? I mistakenly closed the MR

aptalca:
I'll merge it when I'm on a PC, because there are a few other steps I need to do to get it listed in our mod list

PR #810, opened 2024-01-03
Support arbitrary registries, sideloading mods, and multiarch mods

thespad:


Roxedus (approved):

thespad:
@ravensorb incorporated a bunch of your changes into this PR, along with a bunch of other stuff that I've been wanting to do.

ravensorb:
Very nice -- I was just starting to play around with getting the correct arch.

ravensorb:
@thespad question -- I was testing my version and noticed that I needed to switch to using awk when working with apline (when extracting the auth url components).  Did you run into the same issue?

thespad:
The thing you have to be mindful of is that Alpine ships busybox, so common tools like sed, grep, etc. are cut-down compared to the standard GNU varients and sometimes that means options you want aren't available - in this case `grep -P`, so yeah it'll need updating, I tested the code on my host and overlooked the options being used.

We could just add GNU grep to all the Alpine bases (or install it as part of the init if mods are present), but I'd rather avoid doing it just for this.

ravensorb:
Agreed -- I found that this works well

```
    local realm_url=$(echo "$auth_header" | awk -F'[="]+' '/realm=/{print $2}')
    local service=$(echo "$auth_header" | awk -F'[="]+' '/service=/{print $4}')
    local scope=$(echo "$auth_header" | awk -F'[="]+' '/scope=/{print $6}')
```

you can see the working version here: 
https://github.com/ravensorb/docker-mods/blob/81c6ae5a4fc682496d97426af6d55a3afbbf046b/docker-mods.v3#L349

ravensorb:
Question -- I noticed you kept in the hardcoded urls for a few of the registries and was curious -- did I miss a few test cases?

thespad:
Probably 99%+ of our users are going to be pulling from one of those fixed registries, so it's silly to have the script make extra calls to retrieve the auth URL when we already know what it is. Technically it would be *safer* as if the auth URLs change for some reason we'd still capture them, but a) that would break a ton of other people's stuff anyway so it's unlikely to ever happen and b) on the scale we're operating those extra calls add up quickly.

ravensorb:
Ok, that makes sense especially the comment around the extra calls at scale.  Thanks...

PR #823, opened 2024-01-24
add multi-arch support to template

aptalca:
needs https://github.com/linuxserver/github-workflows/pull/46 merged first

### DIFF
--- a/Dockerfile.complex
+++ b/Dockerfile.complex
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:1
 
 ## Buildstage ##
-FROM ghcr.io/linuxserver/baseimage-alpine:3.17 as buildstage
+FROM ghcr.io/linuxserver/baseimage-alpine:3.19 as buildstage
 
 RUN \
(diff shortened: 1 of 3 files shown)


Dockerfile.complex
```diff
@@ -1,17 +1,25 @@
...
 RUN \
   echo "**** install packages ****" && \
   apk add --no-cache \
     curl && \
   echo "**** grab rclone ****" && \
   mkdir -p /root-layer && \
-  curl -o \
-    /root-layer/rclone.deb -L \
-    "https://downloads.rclone.org/v1.47.0/rclone-v1.47.0-linux-amd64.deb"
+  if [[ $(uname -m) == "x86_64" ]]; then \
```
thespad, line 12: I don't think this will work because `[[` is Bash syntax and the Dockerfile runs with `sh` but default

aptalca: Huh, good point. But it seems to work surprisingly. Maybe in alpine it does. I'll change them all.

thespad (approved):

PR #864, opened 2024-03-13
code-server: php-cli initial release

Saij:
code-server: php-cli
Adding php-cli and composer to code-server container


root/etc/s6-overlay/s6-rc.d/init-mod-code-server-php-cli-install/run
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/with-contenv bash
+
+echo "**** Setting up PHP configuration files ****"
+if [ ! -d /config/.php ]; then
+    cp -rf /etc/php /config/.php
+    chown -R abc:abc /config/.php
```
aptalca, line 6: please use our `lsiown` alias instead of `chown`, which works better with remote or fuse mounts

Saij: Is fixed

Issue #789, opened 2023-11-20
problems with Uptime-Kuma mod


apperrault:
hey all
I am running SWAG via docker compose with an override.  I originally got the docker mod to work, and it was great.  Then I made the mistake of removing a container that had been setup to be monitored and now the docker mod is erroring out and stopping swag when i try and restart it.

here is a link to the swag logs from when I try and start it with the mod enabled
https://pastebin.com/1WdwFLpV

here is the pertinent portions of my compose
 `  swag:
    volumes:
      - ${DOCKERCONFDIR}/organizr/www/db:/organizrlog:ro
    environment:
      - MAXMINDDB_LICENSE_KEY=XXXXXXXXXXXXXXX
      - DOCKER_MODS=linuxserver/mods:swag-dashboard|linuxserver/mods:swag-auto-reload|linuxserver/mods:universal-docker|linuxserver/mods:swag-cloudflare-real-ip|linuxserver/mods:swag-auto-proxy #|linuxserver/mods:swag-auto-uptime-kuma
      - DISC_HOOK=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
      - DISC_ME=XXXXXXXXXXXXXXXXXXXXXXXXXXX
      - DOCKER_HOST=tcp://dockerproxy:2375
      - CROWDSEC_API_KEY=XXXXXXXXXXXXXXXXXXXXXXX
      - CROWDSEC_LAPI_URL=http://crowdsec:9091
      - UPTIME_KUMA_URL=http://uptime-kuma:3001/
      - UPTIME_KUMA_USERNAME=admin
      - UPTIME_KUMA_PASSWORD=XXXXXXXXXXXXXXXXXXXXX`
  
Any thoughts on how to fix this?
thanks
app

flowcool:
Same here. I just setted up the mod but had uptime kuma running for a while with existing probe.

```
2023-11-21T17:16:58.841510798Z [mod-auto-uptime-kuma] Deleting all monitors that had their containers removed or were disabled
2023-11-21T17:16:58.843282963Z Traceback (most recent call last):
2023-11-21T17:16:58.843526095Z   File "/app/auto-uptime-kuma.py", line 66, in <module>
2023-11-21T17:16:58.848295867Z     swagUptimeKuma.deleteMonitors(monitorsToBeRemoved)
2023-11-21T17:16:58.848475541Z   File "/app/swagUptimeKuma.py", line 105, in deleteMonitors
[...]

PR #866, opened 2024-03-18
Big updates to Uptime Kuma mod

labmonkey:
This PR pretty much fixes https://github.com/linuxserver/docker-mods/issues/789
But it is also so much more:
- Refactored whole code so that its not such a mess
- Updated documentation so that it explains better how the mod works, requirements, setting up labels and notifications
- Reduced number of API calls when swag container restarts
- Added better support for setting up notifications and fixed issue with default notifications not being applied
- Fixed an issue with mod crashing when there were Manually added monitors in Uptime Kuma
- Fixed an issue with labels that container numeric values
- Added basic support for Monitor Groups
- More log messages so that its easier to grasp what is happening
- Fixed a crash when mod was executed while uptime kuma was not running (mod just gently stops now)

enjoy


labmonkey:
Any chance for this to be reviewed and merged?

aptalca:
Sorry, I saw this when first opened, then totally forgot about it. Reviewing now

.github/workflows/BuildImage.yml
```diff
@@ -7,6 +7,7 @@ env:
   ENDPOINT: "linuxserver/mods" #don't modify
   BASEIMAGE: "swag" #replace
   MODNAME: "auto-uptime-kuma" #replace
+  MULTI_ARCH: "true" #set to false if not needed
```
aptalca, line 10: This should be false as the Dockerfile is not multi-arch and both arches get the same exact files

aptalca (approved):

PR #867, opened 2024-03-25
force create venv if pip packages are to be installed

aptalca:
Currently if python is already installed in the image, we don't create a venv but as more distros are switching to python versions after PEP 668, it is causing errors.

This PR will force create venv if `INSTALL_PIP_PACKAGES` var is defined.

Alpine, noble and debian already do not work without venv when pip packages are defined for install. Jammy currently works, but will break once the images are based on noble.

Without this PR, no images with python installed but no venv will work with the universal package install script to install pip packages once we move to noble.

With this PR, certain python based apps may face issues when the user custom installs pip packages and a venv is force created, but that's a very small risk as the same conditions will cause breakage after jammy --> noble rebase even without this PR.

Tested 2 images that fit that criteria, transmission and deluge, with venv created and they worked fine.

### DIFF
--- a/package-install.v1
+++ b/package-install.v1
@@ -1,11 +1,28 @@
 #!/usr/bin/with-contenv bash
 # shellcheck shell=bash
 
+# Version 1
+# 2024-03-26 - Initial Version
+PKG_SCRIPT_VER="3.20240326"
+
+write_mod_info() {
+    local MSG=$*
+    echo "[pkg-install-init] **** $MSG ****"
+}
+
+write_mod_debug() {
+    local MSG=$*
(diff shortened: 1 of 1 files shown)


thespad (approved):
Can I suggest versioning the script internally like I have with the [mods script](https://github.com/linuxserver/docker-mods/blob/mod-scripts/docker-mods.v3#L8) with a debug option to [print it to stdout](https://github.com/linuxserver/docker-mods/blob/mod-scripts/docker-mods.v3#L297)

It means if you're troubleshooting a container you can easily tell which version of the script is being run.

thespad (approved):

PR #875, opened 2024-04-13
let lsiown ignore broken symlinks

aptalca:
Currently, if lsiown encounters a broken symlink (target not exists), it incorrectly prints the error about remote mounts and such.

This affects SWAG on a fresh install because the certs aren't generated yet.

This PR attempts to ignore broken symlinks in lsiown.

The downside is that the new arg `-xtype` only works with gnu find and not busybox so this PR would require adding the package `findutils` on alpine baseimage.

### DIFF
--- a/docker-mods.v3
+++ b/docker-mods.v3
@@ -5,7 +5,8 @@
 
 # Version 3
 # 2022-09-25 - Initial Release
-MOD_SCRIPT_VER="3.20240225"
+# 2024-04-13 - Let lsiown ignore broken symlinks (requires gnu find)
+MOD_SCRIPT_VER="3.20240413"
 
 # Define custom folder paths
 SCRIPTS_DIR="/custom-cont-init.d"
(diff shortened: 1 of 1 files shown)


quietsy (approved):
Looks good

nemchik:
Do other base images (aside from alpine) also need gnu find packages?

aptalca:
I checked all the other ones and they all have gnu find already. Alpine was the only one with busybox find.

nemchik:
Ref https://unix.stackexchange.com/a/38691

There is an alternative (although it is mentioned to be less efficient than `find . -xtype l`)
We can use `find . -type l ! -exec test -e {} \; -print` (where the `-exec test -e {} \;` is the important part to test if the target exists) which is compatible with busybox find (which is included in all base images already).

For our specific use case the full line would be
```shell
	/usr/bin/find "${PATH[@]}" "${MAXDEPTH[@]}" \( ! -group "${GROUP}" -o ! -user "${USER}" \) -exec test -e {} \; -exec chown "${OPTIONS[@]}" "${USER}":"${GROUP}" {} + || printf "${ERROR}"
```

Still viable to just go ahead with gnu find as well.

nemchik (approved):

PR #890, opened 2024-04-30
transmission-trguing new mod creation

azlux:
Since the project transmission-web-control is archived, I've make a branch for the projet https://github.com/openscopeproject/TrguiNG as a mod.

The base code is from the transmission-transmission-web-control branch.

Best regards,
Azlux


aptalca:
Code needs to be based on the template branch: https://github.com/linuxserver/docker-mods/tree/template?tab=readme-ov-file#mod-creation-instructions

azlux:
> Code needs to be based on the template branch: https://github.com/linuxserver/docker-mods/tree/template?tab=readme-ov-file#mod-creation-instructions

ok, I will rebase it

PR #891, opened 2024-04-30
trguing for transmission: new mod creation

azlux:
Since the project transmission-web-control is archived, I've make a branch for the projet https://github.com/openscopeproject/TrguiNG as a mod.

Best regards,
Azlux


azlux:
Test OK, should be good !

.github/workflows/BuildImage.yml
```diff
@@ -5,8 +5,8 @@ on: [push, pull_request_target, workflow_dispatch]
 env:
   GITHUB_REPO: "linuxserver/docker-mods" #don't modify
   ENDPOINT: "linuxserver/mods" #don't modify
-  BASEIMAGE: "replace_baseimage" #replace
-  MODNAME: "replace_modname" #replace
+  BASEIMAGE: "transmission" #replace
+  MODNAME: "trguing" #replace
   MULTI_ARCH: "true" #set to false if not needed
```
aptalca, line 10: this should be false as the Dockerfile is not arch specific

aptalca (changes requested):
service should depend on `init-mods` instead of `init-mods-package-install` as it doesn't install or need any packages

azlux:
Sorry for all the small commits, I try to understand the usage of mods dependencies/scripts structure.

aptalca:
no worries, it's fine

aptalca (approved):

PR #895, opened 2024-05-12
radarr: striptracks release 2.6.1

TheCaptain989:
- **New** Command-line options can now be passed to the script via the STRIPTRACKS_ARGS environment variable
- **New** Added check for invalid eventtypes
- **New** Allow use of optional environmental variables for Radarr/Sonarr config
- File owner change is only attempted when running as root
- Updated Docker labels


aptalca (approved):
Oops, totally forgot to review this.

Issue #899, opened 2024-05-16
[BUG] Path to FFMPEG No Longer Updatable in 10.8.13


a-priestley:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Name of mod

jellyfin-rffmpeg

### Name of base container

jellyfin

### Current Behavior

As of 10.8.13, this mod is no longer usable due to the inability for users to manually set the path to the ffmpeg binary.

[See here](https://jellyfin.org/posts/jellyfin-security-and-you/)

### Expected Behavior

The way by which the mod operates may have to change. Setting the value in encoding.xml followed by a restart does not seem to work either. Possibly due to:

> Note too that the --ffmpeg flag may or may not change this for you, depending on what version you originally installed; while you should change it to match, you may still need to update the XML file as well.

Note that in the `run` script of `docker-jellyfin`, the flag mentioned above is hard coded, possibly overwriting the value entered into the .xml:

```sh
#!/usr/bin/with-contenv bash
# shellcheck shell=bash

export \
    JELLYFIN_DATA_DIR="/config/data" \
    JELLYFIN_CONFIG_DIR="/config" \
    JELLYFIN_LOG_DIR="/config/log" \
    JELLYFIN_CACHE_DIR="/config/cache" \
    JELLYFIN_WEB_DIR="/usr/share/jellyfin/web"

exec \
    s6-notifyoncheck -d -n 300 -w 1000 \
        s6-setuidgid abc /usr/bin/jellyfin \
        --ffmpeg=/usr/lib/jellyfin-ffmpeg/ffmpeg
```

### Steps To Reproduce

1. Running with the following docker-compose.yml service: 
```yml
services:
  jellyfin:
    image: lscr.io/linuxserver/jellyfin:arm64v8-10.8.13
    network_mode: host
    container_name: jellyfin
    environment:
      - PUID=1000
      - PGID=1000
      - TZ=Etc/UTC
      - JELLYFIN_PublishedServerUrl=192.168.0.120
      - DOCKER_MODS=linuxserver/mods:jellyfin-rffmpeg
      - RFFMPEG_USER=jellyfin
      - RFFMPEG_HOST=100.64.0.1
      - RFFMPEG_HOST_MAC="50:eb:f6:7d:e9:f7"
      - RFFMPEG_WOL=native
      - FFMPEGDIR=/usr/local/bin/wol_rffmpeg/ffmpeg
      - WOL_NATIVE_HOST=192.168.2.101
[...]

PR #900, opened 2024-05-18
Update README

a-priestley:
ffmpeg path is no longer updatable. Edited the mod README to reflect the requirement to set env `FFMPEG_PATH` to override the run flag.


aptalca:
The mod can set the env automatically without rewriting the user to add it. 
See here: https://github.com/linuxserver/docker-mods/blob/transmission-trguing/root/etc/s6-overlay/s6-rc.d/init-mod-transmission-trguing-install/run#L4

a-priestley:
The mod has two possible options for the variable though. Would we default to one with the option of using the other?

aptalca:
Oh I hadn't noticed that. Ignore me then

junkman690:
As discussed in #899 we can probably strip this to the one version, and then have the mod set the variable with a check to see if the user has specified one manually. This would maintain backwards compatibility as no additional user setting need to be made to enable to mod to function for existing installs.

a-priestley:
So we don't need to set the variable if intending to use normal rffmpeg? Only for wol_rffmpeg. I can update the PR if so

junkman690:
I was planning to just auto set it to wol rffmpeg if nothing is set by the user var, and document them both as optional in the readme. Need to add the port etc to readme also from previous pulls. I'm away currently but hoping to get time tonight to quickly add a one line if to set the var to wol if not set by user and update readme.

The wol rffmpeg version is just a sys link to the wol bash script which acts as a wrapper to rffmpeg. If running without wol only addition is a single ping to host and a log entry if the host is down.

Having var auto set as a backup will ensure the mod works as previous for anyone who set ffmpeg manually and isn't keeping up with the release notes.

a-priestley:
superseded by #901

PR #906, opened 2024-05-31
Maxmind: Add new required var for userid

aptalca:
Alpine 3.20 libmaxmind package requires the user id set as well

On earlier versions, this PR should only print a warning, but not actually break anything.

### DIFF
--- a/root/etc/s6-overlay/s6-rc.d/init-mod-swag-maxmind-setup/run
+++ b/root/etc/s6-overlay/s6-rc.d/init-mod-swag-maxmind-setup/run
@@ -12,7 +12,12 @@ if [[ ! -d /var/lib/libmaxminddb ]]; then
     ln -s /config/geoip2db /var/lib/libmaxminddb
 fi
 # check GeoIP2 database
-if [[ -n "$MAXMINDDB_LICENSE_KEY" ]]; then
+if [[ -n "${MAXMINDDB_USER_ID}" ]]; then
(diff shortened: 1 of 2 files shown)


quietsy (approved):

root/etc/s6-overlay/s6-rc.d/init-mod-swag-maxmind-setup/run
```diff
@@ -12,7 +12,12 @@ if [[ ! -d /var/lib/libmaxminddb ]]; then
     ln -s /config/geoip2db /var/lib/libmaxminddb
 fi
 # check GeoIP2 database
-if [[ -n "$MAXMINDDB_LICENSE_KEY" ]]; then
+if [[ -n "${MAXMINDDB_USER_ID}" ]]; then
+    sed -i "s|.*MAXMINDDB_USER_ID.*|MAXMINDDB_USER_ID=\"${MAXMINDDB_USER_ID}\"|g" /etc/libmaxminddb.cron.conf
+else
+    echo "Starting with Alpine 3.20 rebase, MaxMindDB now requires setting the env var MAXMINDDB_USER_ID with your account's user id."
```
golles, line 18: I think Account ID would've been a better name than user id, as that is how it's named in the MaxMind Account Information page

aptalca: That's what alpine uses:
https://git.alpinelinux.org/aports/tree/main/libmaxminddb/libmaxminddb.confd?h=3.20-stable

PR #910, opened 2024-06-06
code-server: add pnpm support

phyesix:
I am making changes according to the new structure and sending it again.


aptalca:
Needs to be based on the template branch, not master

PR #916, opened 2024-06-11
Add support for multiple docker hosts

FoxxMD:
* generate configs using both docker.sock and DOCKER_HOST data
* parse multiple hosts from DOCKER_HOST
* enable generating swag_url with host friendly name inserted
* enable setting TLD globally or per-host
* maintain backwards compatibility with current functionality

Replicated from the updated readme:

## Multiple Hosts:

If both `DOCKER_HOST` and `docker.sock` volumes are provided this mod will detect containers using both connections. As noted in the [requirements](#requirements), containers detected via `docker.sock` must be in the same user defined network or have `swag_address` label set.

Multiple remote hosts can be used via `DOCKER_HOST` by separating hosts with a comma. Additional per-host settings can be assigned by separating with a pipe `|`. The syntax for per-host configuration:

```
host:port|friendly_name|default_tld
```

```
DOCKER_HOST=192.168.0.100:2375|serverA,192.168.0.110:2375|serverB|local.test,192.168.0.130:2375
```

* Host: `192.168.0.100:2375` -- Friendly Name: `serverA` -- TLD: `*`
* Host: `192.168.0.110:2375` -- Friendly Name: `serverB` -- TLD: `local.test`
* Host: `192.168.0.130:2375` -- Friendly Name: `host3` -- TLD: `*`

### Upstream IP and Port

When using a remote docker host from `DOCKER_HOST` auto-proxy assumes the detected containers are not on the same network as SWAG:

* If the detected containers do not have the `swag_address` label set then the Host IP will be used.
* If the detected containers do not have the `swag_port` label set then auto-proxy looks for exposed **container ports** and uses the corresponding **host port** as the upstream port. Container ports are checked in this order:
  * 80
  * 8080
  * The first mapped port, if any

### Subdomains and TLD

If a detected container does not have the `swag_url` label set then the subdomain and TLD can be programmatically generated.

The default TLD used in nginx [`server_name` directive](https://nginx.org/en/docs/http/server_names.html) can be set using `AUTO_PROXY_HOST_TLD`. This can also be set per-host using the syntax described in [`DOCKER_HOST` for `default_tld`.](#multiple-hosts)

The subdomain used for a container can optionally be modified to include the Host's `friendly_name` described in the `DOCKER_HOST` syntax by setting `AUTO_PROXY_HOST_INSERT` to either `prefix` or `suffix`

Examples using a container named `overseer`:

* Using only AUTO_PROXY_HOST_INSERT to modify subdomain
  * `DOCKER_HOST=192.168.0.100:2375|serverA`
  * `AUTO_PROXY_HOST_TLD` (not set, defaults to `*`)
  * `AUTO_PROXY_HOST_INSERT`
    * (unset) => nginx `server_name overseer.*`
    * `prefix` => nginx `server_name serverA-overseer.*`
    * `suffix` => nginx `server_name overseer-serverA.*`
* Using AUTO_PROXY_HOST_INSERT prefix and AUTO_PROXY_HOST_TLD
  * `DOCKER_HOST=192.168.0.100:2375|serverA`
  * `AUTO_PROXY_HOST_TLD=test.home`
  * `AUTO_PROXY_HOST_INSERT=prefix`
    * `server_name serverA-overseer.test.home`
* Using AUTO_PROXY_HOST_INSERT prefix and default_tld
  * `DOCKER_HOST=192.168.0.100:2375|serverA|myserver.home`
  * `AUTO_PROXY_HOST_INSERT=prefix`
    * `server_name serverA-overseer.myserver.home`
* Using AUTO_PROXY_HOST_TLD only
  * `DOCKER_HOST=192.168.0.100:2375`
  * `AUTO_PROXY_HOST_TLD=myserver.home`
    * `server_name overseer.myserver.home`

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,23 +41,58 @@ Then the env var in SWAG can be set as `DOCKER_HOST=dockerproxy`. This will allo
 
 If both `DOCKER_HOST` and `docker.sock` volumes are provided this mod will detect containers using both connections. As noted in the [requirements](#requirements), containers detected via `docker.sock` must be in the same user defined network or have `swag_address` label set.
 
-Multiple remote hosts can be used via `DOCKER_HOST` by separating hosts with a comma. Additionally, a friendly host name can be assigned by suffixing the host with `|friendlyName`. If the host is not assigned a friendly name it will have a default generated like `hostN` where N is the position of the host in `DOCKER_HOST`. Example:
+Multiple remote hosts can be used via `DOCKER_HOST` by separating hosts with a comma. Additional per-host settings can be assigned by separating with a pipe `|`. The syntax for per-host configuration:
 
 ```
-DOCKER_HOST=192.168.0.100:2375|serverA,192.168.0.110:2375|serverB,192.168.0.130:2375
+host:port|friendly_name|default_tld
 ```
 
-* Host: `192.168.0.100:2375` -- Friendly Name: `serverA`
-* Host: `192.168.0.110:2375` -- Friendly Name: `serverB`
-* Host: `192.168.0.130:2375` -- Friendly Name: `host3`
 
-If the detected containers for each host do not have the `swag_address` label set then the host IP will be used.
+```
+DOCKER_HOST=192.168.0.100:2375|serverA,192.168.0.110:2375|serverB|local.test,192.168.0.130:2375
+```
 
-If the detected containers for each host do not have the `swag_url` label set then the default can be configured with friendly host name inserted into the url as either a suffix or prefix using the env `HOST_INSERT`. Example:
+* Host: `192.168.0.100:2375` -- Friendly Name: `serverA` -- TLD: `*`
+* Host: `192.168.0.110:2375` -- Friendly Name: `serverB` -- TLD: `local.test`
+* Host: `192.168.0.130:2375` -- Friendly Name: `host3` -- TLD: `*`
 
-* container name: `swag`
-* `DOCKER_HOST=192.168.0.100:2375|serverA`
-* `HOST_INSERT`
-  * (unset) => nginx `server_name swag.*`
-  * `prefix` => nginx `server_name serverA-swag.*`
-  * `suffix` => nginx `server_name swag-serverA.*`
+### Upstream IP and Port
+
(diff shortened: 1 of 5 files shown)


FoxxMD:
Other than needing to fast-forward with the current tree there is no other feedback I have received. The changes are feature complete as far as I'm concerned. I've been using this on my own servers for weeks. I guess I'll mention some previous contributors since no one is looking at this? @aptalca @nemchik ?

FoxxMD:
The PR is still ready other than needing to be synced with the upstream. Can someone please take a look?

aptalca:
Thanks for the PR, but unfortunately this is not something I (the maintainer) can take on in terms of support and future maintenance.

The changes are very extensive, and it even adds a wrapper while the benefit is not that great. I do see the benefit, but very few people would actually need this, which makes this a niche feature. 

I'm afraid the cost to benefit ratio is way too high for me.

I'll leave it open in case another team member is interested in taking over maintenance and is ok with merging this in.

dacagi:
I would actually love to have this feature. I have been most of the day looking for a way to autodiscover from my two hosts. Most people may start with running on a single Raspberry, but once you start to self-host there's no stop. It's worse than Pringles.

I understand the whole project is about a single host, but it has already grown to even configuring Uptime. Why not reaching out to another host on the network? TCP communication instead of only socket is already available.

I just arrived and don't know much about the internals, but this would be a nic(h)e feature. I haven't found any other project that allows this without getting into container orchestration swarms and so on. It would really make SWAG shine.

Thanks FoxxMD for the huge work, I may give your version a try.

FoxxMD:
Just FYI my fork can be used even though this isn't merged. I've pushed an image to docker (or the LSIO pr image could be used):

```yaml
services:
  swag:
   #...
    environment:
      DOCKER_MODS: "linuxserver/mods:universal-docker|foxxmd/auto-proxy-multi"
```

FoxxMD:
I plan to create a new project that will achieve the same functionality with a bit more of an agnostic platform. I'll comment here once I've got it started. My fork and published image are still working with current SWAG and should be usable for the foreseeable future.

PR #919, opened 2024-06-18
code-server-php-ppa initial release

aptalca:
This is a more expansive version of the php8 mod, which is hardcoded to php8.2

Once this is merged, we can deprecate the php8 mod and refer users to this one.


aptalca:
Ugh, after all this work, I realized we already have a very similar one: https://github.com/linuxserver/docker-mods/tree/code-server-php-cli

Closing

Issue #948, opened 2024-09-02
[BUG] Emby fails to start with Emby-Mediainfo-Plugin Mod


DaveO16176:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Name of mod

emby-mediainfo-plugin

### Name of base container

Emby

### Current Behavior

When the Emby container restarts the mod forces the container to exit because the unzip command queries for user input to overwrite the existing biftool_linux.zip file.  Initially creating the container or recreating the container avoids this error.

### Expected Behavior

The mod should suppress the overwrite query and overwrite the existing biftool_linux.zip file if the mod downloaded a newer version of the file.

### Steps To Reproduce

1. Restart an Emby container with the emby-mediainfo-plugin as a docker mod
2. Container's log will show the unzip command querying for user input and eventually existing, which will cause the Emby container to exit

### Environment

```markdown
- OS:
- How docker service was installed:
```

### CPU architecture

x86-64

### Docker creation

```bash
version: '3.8'
services:
  emby:
    image: lscr.io/linuxserver/emby:latest
    container_name: emby
    network_mode: host
    environment:
      - PUID=1000
      - PGID=1000
      - TZ=America/New_York
      - NVIDIA_VISIBLE_DEVICES=all
      - DOCKER_MODS=linuxserver/mods:emby-mediainfo-plugin
    runtime: nvidia
    volumes:
      - /media/emby/config:/config
      - /media/pool:/mnt/share1
      - /media/a_star_hdd/recordings:/recordings
      - /media/a_star_hdd/transcode:/transcode
    ports:
      - 8096:8096
    restart: unless-stopped
```

### Container logs

```bash
2024-09-02T17:56:30.244798329Z [mod-init] Running Docker Modification Logic
2024-09-02T17:56:30.410786139Z [mod-init] Adding linuxserver/mods:emby-mediainfo-plugin to container
2024-09-02T17:56:31.185342424Z [mod-init] linuxserver/mods:emby-mediainfo-plugin at sha256:1f90339379c476cd40b6a4badb7147dd9677f1825b03e2823473fa9128d1b8cc has been previously applied skipping
2024-09-02T17:56:31.213295435Z [migrations] started
2024-09-02T17:56:31.213320552Z [migrations] no migrations found
2024-09-02T17:56:31.231177001Z usermod: no changes
2024-09-02T17:56:31.231943326Z ───────────────────────────────────────
2024-09-02T17:56:31.231958605Z 
[...]

PR #949, opened 2024-09-02
emby: mediainfo-plugin fix unzip command

DaveO16176:

 - [X ] I have read the [contributing](https://github.com/linuxserver/docker-mods/blob/main/.github/CONTRIBUTING.md) guideline and understand that I have made the correct modifications

## Description:
Adds command options to the unzip command when the mod is checking to see if it should install and overwrite the biftool_linux.zip file.

Fixes the issue documented in [Issue #948].  Without this bug fix, users are forced to recreate the Emby container so the mod successfully runs.

## How Has This Been Tested?
Built a local version with the new unzip options and the docker container successfully restarts.

### DIFF
--- a/root/etc/s6-overlay/s6-rc.d/init-mod-emby-mediainfo-plugin-postinstall/run
+++ b/root/etc/s6-overlay/s6-rc.d/init-mod-emby-mediainfo-plugin-postinstall/run
@@ -1,12 +1,12 @@
 #!/usr/bin/with-contenv bash
 
-# This is an install script that is designed to run after init-mods-package-install
-# so it can take advantage of packages installed
-# init-mods-end depends on this script so that later init and services wait until this script exits
-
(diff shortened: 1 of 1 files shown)


aptalca:
@DaveO16176
I made the BIF Tool install conditional
Please test the built PR and confirm it works
Thanks

DaveO16176 (commented):
@aptalca I tested the built PR and it works fine after the container is restarted.

I had originally thought of making the installation of the bif tool conditional like you did; however, if Roku ever updates the bif tool then the new version would never get downloaded, which is why I just added the -uo option to unzip. Roku hasn't updated the executable in 5 years so I don't think it's likely, but there could be a chance.

aptalca:
5 years is eternity. If it ever gets updated, we'll make a change then. No need to redownload it every time the container starts until then.

PR #951, opened 2024-09-11
Bundle binaries at buildtime

thespad:

### DIFF
--- a/root/etc/s6-overlay/s6-rc.d/init-mod-universal-calibre-setup/run
+++ b/root/etc/s6-overlay/s6-rc.d/init-mod-universal-calibre-setup/run
@@ -17,18 +17,11 @@ fi
 
 export DEBIAN_FRONTEND="noninteractive"
 
-CALIBRE_RELEASE="$(cat /CALIBRE_RELEASE)"
-if [ -z ${CALIBRE_RELEASE+x} ]; then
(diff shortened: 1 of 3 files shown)


root/etc/s6-overlay/s6-rc.d/init-mod-universal-calibre-setup/run
```diff
@@ -35,29 +21,14 @@ CALIBRE_RELEASE="$(cat /CALIBRE_RELEASE)"
...
     tar xf \
         /tmp/calibre.txz \
         -C /app/calibre
     rm /tmp/calibre.txz
+    echo "Installing Calibre version $(cat /CALIBRE_RELEASE)"
     /app/calibre/calibre_postinstall
     echo "${CALIBRE_RELEASE}" > /config/.CALIBRE_RELEASE
-else
-    echo "**** Calibre already installed, skipping ****"
 fi
```
aptalca, line 34: I think this logic would fail if the user restarts the container after a new calibre version is released, but the mod is not built yet.

Might be simpler to just attempt an install if `/tmp/calibre.txz` exists, which would suggest a mod update and after install it is deleted so restarts with no mod update would skip it.

thespad: Done.

aptalca (approved):

PR #969, opened 2024-10-19
Move contenv alias to build time for future bases

thespad:


Roxedus:
Do we want to add a mention for the earliest potential removal of the dynamic creation of contenv? ie next LTS of ubuntu

thespad:
In principle we could backdate the change at any point, I'm just not hugely concerned about doing it because there's no massive benefit right now.

Roxedus (approved):

PR #973, opened 2024-11-03
Adding documentation regarding legacy CPU deprecation in Intel's compute runtime

solidsnake1298:
Adds documentation regarding Intel deprecating pre-Tigerlake iGPUs for their compute runtime.


aptalca (approved):
Looks great. Thanks for the PR.

PR #976, opened 2024-11-11
Update dashboard.subdomain.conf.sample

hellblazer315:
Changed from explicit IP ranges to using an included config file. Recommend this shouldn't be merged upstream unless [https://github.com/linuxserver/docker-swag/pull/520](https://github.com/linuxserver/docker-swag/pull/520) is as well.

## Description:
Removed all explicit IP Address ranges and instead added an include to a static config file.

**Dependency**: Requires an internal.conf file as proposed by [https://github.com/linuxserver/docker-swag/pull/520](https://github.com/linuxserver/docker-swag/pull/520)

Allows a single place to manage IPs instead of 3 different places in the file. Also allows easier re-use of this configuration in other proxy configs.

## How Has This Been Tested?
Created the internal.conf from https://github.com/linuxserver/docker-swag/pull/520 then removed all IP blocks in dashboard.subdomain.conf and replaced with "include /config/nginx/internal.conf". Confirmed access is still restricted to local IPs.

## Source / References:
[https://github.com/linuxserver/docker-swag/pull/520](https://github.com/linuxserver/docker-swag/pull/520)

### DIFF
--- a/root/dashboard/dashboard.subdomain.conf.sample
+++ b/root/dashboard/dashboard.subdomain.conf.sample
@@ -34,10 +34,8 @@ server {
         # enable for Authentik (requires authentik-server.conf in the server block)
         #include /config/nginx/authentik-location.conf;
 
-        allow 10.0.0.0/8;
-        allow 172.16.0.0/12;
-        allow 192.168.0.0/16;
-        deny all;
+        # Only permit access from internal IP Addresses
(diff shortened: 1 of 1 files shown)


quietsy:
As we've discussed on discord, this will break backward compatibility with older versions of swag.

PR #988, opened 2024-12-14
switch to lsio qemu-static

aptalca:


root/etc/s6-overlay/s6-rc.d/init-mod-universal-docker-in-docker-qemu/run
```diff
@@ -7,5 +7,6 @@ fi
 
 if [[ $(uname -m) = "x86_64" ]]; then
```
thespad, line 8: Could remove this now our qemu image supports use in both directions.

aptalca: Good point 👍

thespad (approved):

PR #1021, opened 2025-03-23
Update the Opencl-Intel mod from 24.35.30872.22 to the latest fix 24.35.30872.32

lluisd:
## Description:
There is that legacy 24.35 branch for legacy releases [which the last fix is that ](https://github.com/intel/compute-runtime/commits/releases/24.35/) which the latest fix is the 24.35.30872.32, so use that one istead of the 24.35.30872.22 current one.

This is from the description of the 24.35 branch of the previous link which they say that at the end of the readme.file https://github.com/intel/compute-runtime/blob/master/LEGACY_PLATFORMS.md

## Source / References:
legacy description https://github.com/intel/compute-runtime/blob/master/LEGACY_PLATFORMS.md
24.35 legacy branch https://github.com/intel/compute-runtime/commits/releases/24.35/
https://github.com/intel/compute-runtime/releases/tag/24.35.30872.32

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN \
   if [ -z "${MOD_VERSION}" ]; then \
     MOD_VERSION=$(curl -sX GET "https://api.github.com/repos/intel/compute-runtime/releases/latest" | jq -r '.tag_name'); \
   fi && \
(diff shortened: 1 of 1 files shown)


aptalca (approved):

aptalca:
It would be nice if this PR was tested by the op. Upstream tagged it but did not cut a release so our build is unable to pick up any artifacts.

Will have to revert.

PR #1030, opened 2025-05-16
Instructions missing config in dbip.conf

pgherkin:
The example dbip.conf configuration should include: Setting lan-IPs and loading ISO country codes to avoid [emerg] errors

 - [ ] I have read the [contributing](https://github.com/linuxserver/docker-mods/blob/main/.github/CONTRIBUTING.md) guideline and understand that I have made the correct modifications

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,11 @@ Follow these steps to enable the dbip mod:
 4. Edit `/config/nginx/dbip.conf` and add countries to the blocklist / whitelist according to the comments, for example:
    
     ```nginx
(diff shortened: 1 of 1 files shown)


quietsy:
The [example config](https://github.com/linuxserver/docker-mods/blob/swag-dbip/root/defaults/dbip.conf) does include it, the instructions clearly state that users should `add countries to the blocklist / whitelist according to the comments` with an example below, the example isn't supposed to include the entire config.

Thanks for your contribution but I don't feel it's necessary to over-complicate the instructions.

PR #1050, opened 2025-07-22
Fix swag dashboard

quietsy:
Fix swag dashboard due to changes in goaccess css


thelamer:
I do not use this mod, but I trust you have tested this properly and the code looks logical.

PR #1075, opened 2025-11-11
update upstream release tag

aptalca:


aptalca:
Ignore the PR build failure. It fails because the PR build ignores the fix in BuildImage.yml

PR #1159, opened 2026-07-03
Swag ondemand fix websockets

### LINKED ISSUE #1090: [BUG] swag-ondemand: Ignores websockets
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Name of mod

When using something like linuxserver/firefox with swag_ondemand, swag will kill the container while it is being used, because all traffic is done via an open websocket. I guess that is because nothing is logged as the connection is long living.

### Name of base container

swag

### Current Behavior

Kills active container

### Expected Behavior

The container should not be killed while in use

### Steps To Reproduce

1. Start linuxserver/firefox with swag_ondemand
2. Use it
3. Wait for the kill timeout

### Environment

```markdown
- OS:
- How docker service was installed:
```

### CPU architecture

x86-64

### Docker creation

```bash
.
```

### Container logs

```bash
2025-11-29 22:07:27 - Thread-1 - INFO - Started firefox
2025-11-29 22:23:43 - Thread-1 - INFO - Stopped firefox after 600s of inactivity
2025-11-29 22:23:48 - Thread-1 - INFO - Started firefox
```

quietsy:
Set `swag_ondemand_websocket=1` on apps that use a websocket to communicate, to prevent ondemand from stopping them until the browser tab is closed by waiting for a 101 error code in the access log.

Fixes https://github.com/linuxserver/docker-mods/issues/1090

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ Or set the following label if using `swag-auto-proxy`:
 ### Labels:
 - `swag_ondemand=enable` - required for on-demand.
 - `swag_ondemand_urls=https://wake.domain.com,https://app.domain.com/up` - *optional* - overrides the monitored URLs for starting the container on-demand. Defaults to using the value of the `swag_url` label, if you've already set it for `swag-auto-proxy`, or `https://somecontainer.,http://somecontainer.` otherwise.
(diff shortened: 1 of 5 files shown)


aptalca (approved):

aptalca:
tested PR build, works great

PR #1162, opened 2026-07-23
Update index.php

santacruz-usa:
Modified call to goaccess: instead of sending the entire log content to goaccess via stdin, let goaccess process the log file directly.

 - [X] I have read the [contributing](https://github.com/linuxserver/docker-mods/blob/main/.github/CONTRIBUTING.md) guideline and understand that I have made the correct modifications

## Description:

Modified call to goaccess: instead of sending the entire log content to goaccess via stdin, let goaccess process the log file directly.  This change allows goaccess to use built-in efficiencies for parsing: by using a file, goaccess will persist statistics regarding any already-parsed data from the file. Using stdin instead means that goaccess must rely on timestamp comparisons and discard rows that are earlier than the latest parse timestamp. As the log file grows (until logrotation), we can avoid sending already-parsed data repeatedly for every call by having goaccess keep track of its own state.

The benefit of accepting this PR is reduced overhead and improved processing efficiency for the dashboard.

## How Has This Been Tested?

Testing was performed by using docker compose with bind mount for "/custom-cont-init.d", in which a custom script modified 'index.php' to reflect the intended change:
<pre> 
declare PHPFILE="/dashboard/www/index.php"
if [ ! -f "${PHPFILE}" ]; then
    echo "**** '${PHPFILE}' is not in default location; skipping ****"
    exit 0
fi
sed -i -r 's/(shell_exec\(")(cat \$access_log \| )(\/usr\/bin\/goaccess)(.*asndb)( -)/\1\3 $access_log\4/' "${PHPFILE}"
</pre>
For both the original code and the proposed code, dashboard load timings were captured for a series of increasingly larger 'access.log' files. The proposed code showed an improvement in processing efficiency.

Requesting a maintainer review and approve test suite runner (getting pull_request_target error)

## Source / References:

See the documentation for goaccess here, specifically the section "Processing Logs Incrementally": https://goaccess.io/man#examples

### DIFF
--- a/root/dashboard/www/index.php
+++ b/root/dashboard/www/index.php
@@ -291,7 +291,7 @@
         endif;
 
         $access_log = file_exists("/dashboard/logs") ? "/dashboard/logs/*.log" : "/config/log/nginx/access.log";
-        $goaccess = shell_exec("cat $access_log | /usr/bin/goaccess -a -o html --config-file=/dashboard/goaccess.conf $geodb $asndb -");
+        $goaccess = shell_exec("/usr/bin/goaccess $access_log -a -o html --config-file=/dashboard/goaccess.conf $geodb $asndb");
         $goaccess = str_replace("<div class='loading-container'>", "<div hidden>", $goaccess);
         $goaccess = str_replace("<title>Server&nbsp;Statistics", "<title>SWAG&nbsp;Dashboard", $goaccess);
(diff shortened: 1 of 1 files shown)


santacruz-usa:
Kindly requesting a review of this PR to move forward.

quietsy:
Thanks for the contribution, we'll look into it.

quietsy:
I'm seeing a slower load time with this change, what are your results?

